PR #2, opened 2015-04-02
Update the Composer instructions

stof:
These instructions avoid people to make manual edition of the composer.json file.

Once the library is stable, you can even remove the version entirely and Composer will guess the best semver constraint

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,20 +56,10 @@ foreach ($result as $row) {                       // results and rows implement 
 **NOTE**: You must first [install and enable the `cassandra.so` PHP extension](ext/README.md).
 
 * First, [get composer](https://getcomposer.org/).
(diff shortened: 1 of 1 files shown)


avalanche123:
Thanks @stof!

I will also need your help with reviews and feedback on this driver in the future.
Can you please sign the CLA so I can merge this in?

Cheers!

stof:
@avalanche123 what would be the way to discuss with you about some initial feedback ? Twitter is not the appropriate format

avalanche123:
@stof irc freenode #datastax-drivers, or a Jira issue

avalanche123:
oh, also [the mailing list](https://groups.google.com/a/lists.datastax.com/forum/#!forum/php-driver-user). That might be a more appropriate forum.

PR #5, opened 2015-04-02
make SSLOptions immutable

avalanche123:


src/Cassandra/SSLOptions.php
```diff
@@ -24,29 +24,83 @@
...
+    /**
+     * Creates ssl context resource
+     * @access private
+     * @return resource ssl context
+     */
+    public function resource()
+    {
+        $resource = cassandra_ssl_new();
+
+        if (is_array($this->trustedCerts)) {
```
stof, line 86: what about typehinting it as array to avoid this check ? `null` is currently strictly equivalent to an empty array, so you could just use an empty array

src/Cassandra/SSLOptions.php
```diff
@@ -24,29 +24,83 @@
...
+    /**
+     * Constructs SSLOptions instance
+     * Don't instantiate directly, use Cassandra\SSLOptions\Builder instead.
+     *
+     * @access private
+     * @param array|null $trustedCerts Trusted certificate paths
+     * @param int|null   $verifyFlags  A bitmask of Cassandra::VERIFY_* constants
+     * @param string     $clientCert   Client cert
+     * @param string     $privateKey   Path to private key
+     * @param string     $passphrase   Passphrase for the private key if any
```
stof, line 66: missing `|null` for all of these

src/Cassandra/SSLOptions.php
```diff
@@ -24,29 +24,83 @@
...
+    private $verifyFlags = null;
+
+    /**
+     * Constructs SSLOptions instance
+     * Don't instantiate directly, use Cassandra\SSLOptions\Builder instead.
+     *
+     * @access private
+     * @param array|null $trustedCerts Trusted certificate paths
+     * @param int|null   $verifyFlags  A bitmask of Cassandra::VERIFY_* constants
+     * @param string     $clientCert   Client cert
```
stof, line 64: `Path to client sert`

avalanche123:
you're right, fixing these now

src/Cassandra/SSLOptions.php
avalanche123, line 86: good point I'd rather skip an array allocation altogether though

src/Cassandra/SSLOptions.php
```diff
@@ -24,29 +24,83 @@
...
+     * Client cert
+     * @var string|null
+     */
+    private $clientCert = null;
+
+    /**
+     * Path to private key, this is used to verify client's identity.
+     * @var string|null
+     */
+    private $privateKey = null;
```
stof, line 43: `= null` is useless as it is already the PHP behavior when you don't define anything.

avalanche123, line 43: kk, will remove, makes little sense when there is a constructor anyway

PR #6, opened 2015-04-02
Remove an unused property

stof:


avalanche123:
that one is required for the garbage collector to not destroy an ssl context prematurely

avalanche123:
keep these coming

stof:
@avalanche123 this seems like a misconception of the PHP extension though. If the extension itself keeps a reference to the object, it should increment the refcount for that instead of requiring to keep a userland reference

avalanche123:
Fair point, can you file a Jira please?

On Thu, Apr 2, 2015 at 1:31 PM, Christophe Coevoet <notifications@github.com

> wrote:
> 
> @avalanche123 https://github.com/avalanche123 this seems like a
> misconception of the PHP extension though. If the extension itself keeps a
> reference to the object, it should increment the refcount for that instead
> of requiring to keep a userland reference
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/datastax/php-driver/pull/6#issuecomment-89037084.

## 

_Bulat Shakirzyanov_ | Software Alchemist

_a: *about.me/avalanche123
*e:_ mallluhuct@gmail.com

stof:
done: https://datastax-oss.atlassian.net/projects/PHP/issues/PHP-5

avalanche123:
awesome, thanks!

PR #18, opened 2015-04-10
Add cassandra_cluster_set_connect + request_timeout()

jbaldwin:
Tried calling these two functions since they are in the php code, but apparently the php-extension calls don't exist yet but the cpp-driver does have the equivalent.

I also changed the php code to accept int's in milliseconds instead of floats since the cpp-driver use an unsigned int and not a float so there is no conversion required.

I apparently also wasn't pulling the &cluster_resource properly in the port last pull request.  Oops.


avalanche123:
Thanks for this!
To keep things simple I wanted all settings to accept timeouts in seconds or fractions of a second.
I think we still want the builder to accept floats (seconds) it should just multiply the float value by 1000, ceil it and use that internally

jbaldwin:
Ok, I'll amend it to still accept floats then.  Thanks for the feedback.

avalanche123:
sounds good, thanks for your contributions!

PR #19, opened 2015-04-15
Added support for persistent sessions.

jbaldwin:
I'd like to open this up for comments and suggestions and/or changes.

Just as a bit of background this has been deployed to about a dozen of our servers for a day now as a test against live traffic and this iteration of the code hasn't shown any issues yet for us.  I went ahead and added the persistent sessions because creating and tearing down connections to the cassandra cluster per request was too much overhead for our application and caused a significant slowdown.

### DIFF
--- a/ext/php_cassandra.c
+++ b/ext/php_cassandra.c
@@ -155,6 +155,7 @@ const zend_function_entry cassandra_functions[] = {
   PHP_FE(cassandra_ssl_set_verify_flags, NULL)
   /* CassSession */
   PHP_FE(cassandra_session_new, NULL)
+  PHP_FE(cassandra_session_pnew, NULL)
   PHP_FE(cassandra_session_free, NULL)
   PHP_FE(cassandra_session_connect, NULL)
   PHP_FE(cassandra_session_connect_keyspace, NULL)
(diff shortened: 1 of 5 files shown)


avalanche123:
This is a great first step, thanks!

Ultimately we will want this configure via `Cluster\Builder::withPersistentSessions()`.
And we hash by builder settings automatically.
I am in fact working on this very feature.
But I can hold off if you want to contribute it yourself.

jbaldwin:
I would think this would return a new PersistentCluster and possibly
PersistentSession when you call build() and connect()?  I like this but
didn't really think of implementing it this way originally.  Also I don't
think the cluster needs to be persistent, only the session really needs to
be, thoughts?  Maybe there is a new persistent cluster object but the php
extension still only persists the session between requests.

avalanche123:
I'm actually defining Builder and Cluster (and session and ssl options) classes in C, this lets me return the same DefaultCluster class in both scenarios.

The cluster object needs to outlive sessions. While it may not be causing problems today, this is something the C/C++ driver's API assumes.

Bulat

jbaldwin:
Ok, this sounds like you are making some huge changes to how the codebase works (I know it is still alpha so it is all good...).  I think we will continue to use this for the time being and upgrade once you have finished your changes.

Also, if anything in this pull request is useful for your work you can absolutely use it, if not then no harm, I just thought it might be useful.

avalanche123:
Sounds good, keep your contributions coming!

Bulat

avalanche123:
closing in favor of work included in PHP-1

PR #20, opened 2015-04-21
Add null value support

bublikOff:
Text type with null gives me "Segmentation fault" when i tries to get its value;


avalanche123:
you want to change the pull request base to be against branch PHP-12

avalanche123:
Actually, I don't understand what you're submitting, branch PHP-12 is a work in progress, use the master as that is the latest stable.

PR #22, opened 2015-04-29
added logging to file/syslog

avalanche123:


ext/php_cassandra.c
```diff
@@ -173,12 +174,67 @@ ZEND_GET_MODULE(cassandra)
...
-  fprintf(stderr, "php-driver | [%s] %s (%s:%d)\n",
-    cass_log_level_string(message->severity), message->message,
-    message->file, message->line
-  );
+  char log[MAXPATHLEN + 1];
+  uint log_length = 0;
+
+  (void)data;
+
+  uv_rwlock_rdlock(&CASSANDRA_G(log_lock));
```
avalanche123: is this necessary when not threaded?

ext/php_cassandra.c
```diff
@@ -173,12 +174,67 @@ ZEND_GET_MODULE(cassandra)
...
+php_cassandra_log(const CassLogMessage* message, void* data)
 {
-  fprintf(stderr, "php-driver | [%s] %s (%s:%d)\n",
-    cass_log_level_string(message->severity), message->message,
-    message->file, message->line
-  );
+  char log[MAXPATHLEN + 1];
+  uint log_length = 0;
+
+  (void)data;
```
avalanche123: what is this line for?

ext/php_cassandra.c
```diff
@@ -8,6 +9,8 @@
 #include "types/map.h"
 #include "types/set.h"
 
+#define PHP_CASSANDRA_DEFAULT_LOG "php-driver.log"
```
avalanche123, line 12: Will this file be located in the php log directory by default?

ext/php_cassandra.c
```diff
@@ -173,12 +174,67 @@ ZEND_GET_MODULE(cassandra)
...
+  }
+  uv_rwlock_rdunlock(&CASSANDRA_G(log_lock));
+
+  log[log_length] = '\0';
+
+  if (log_length > 0) {
+    int fd = -1;
+
+    if (!strcmp(log, "syslog")) {
+      php_syslog(LOG_NOTICE, "php-driver | [%s] %s (%s:%d)",
```
avalanche123, line 222: Should we do something like http://lxr.php.net/xref/PHP_5_3/sapi/fpm/fpm/zlog.c#185 to map cpp-driver log level to syslog severity?

ext/php_cassandra.c
```diff
@@ -173,12 +174,67 @@ ZEND_GET_MODULE(cassandra)
...
+              PHP_EOL);
+
+      write(fd, tmp, needed);
+      free(tmp);
+      close(fd);
+      return;
+    }
+  }
+
+  fprintf(stderr, "php-driver | [%s] %s (%s:%d)%s",
```
avalanche123, line 264: I remember discussing defaulting to sapi's log function, any reason not to?

avalanche123, line 264: I guess thread-safety would be a concern. We could default to sapi if threading not enabled and write to stderr otherwise, what do you think?

ext/php_cassandra.c
avalanche123: looks like this fails in zts environment, because we're missing `TSRM_CC` in this context, we should use `TSRMLS_SET_CTX` and `TSRMLS_FETCH_FROM_CTX` to pass the thread context via `data` argument

avalanche123:
looks good, once we address my comments

ext/php_cassandra.c
mpenick: The log callback is always called on another thread so it's always needed. There's nothing in here that uses TSRM.

ext/php_cassandra.c
mpenick: To avoid unused warning for the parameter "data".

ext/php_cassandra.c
avalanche123: I believe `CASSANDRA_G` macro does use TSRM

avalanche123: Specifically, here is the error from travis - https://travis-ci.org/datastax/php-driver/jobs/60564028#L553

ext/php_cassandra.c
mpenick, line 264: There are no guarantees that underlying SAPI's log implementation is thread-safe.

ext/php_cassandra.c
avalanche123: And the macro definition - https://github.com/datastax/php-driver/blob/master/ext/php_cassandra.h#L213

ext/php_cassandra.c
avalanche123, line 264: that's true, maybe add a comment about it...

ext/php_cassandra.c
mpenick, line 222: I think the driver's severity != syslog's severity because there are more critical events that could happen on a system and you might not want php-driver stuff showing up in your error/critical level system logs. I could either way here, but I'm in favor of leaving as notice.

avalanche123, line 222: Gotcha, makes sense, let's keep it as is

ext/php_cassandra.c
mpenick: virtual_open() uses it.

avalanche123: yep, good find!

ext/php_cassandra.c
mpenick, line 12: It's in the current working directory (could be the VCWD too). I think for something like "error_log()" the SAPI determines the directory where the PHP logs go. I don't think we have access to that information here.

mpenick:
This will work for now and will work well. The main issue is that under ZTS several threads could be fighting to set the same globals settings, "cassandra.log" and "cassandra.log_level"  The long term solution is to allow a per-session logging callback in the cpp-driver.

avalanche123:
looks good to me, merging

PR #23, opened 2015-05-01
PHP-1

avalanche123:
Windows build support


.gitignore
```diff
@@ -5,3 +5,4 @@
 /.vagrant/
 /venv/
 /tmp/*
+/*.swp
```
avalanche123: is this a vim swap file being ignored, if so, it should be included in the global gitignore on your system - https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

ext/config.m4
```diff
@@ -7,28 +7,92 @@ if test -z "$PHP_GMP"; then
...
+
+  case $(uname -s) in
+    Linux)
+      CASSANDRA_CFLAGS="-Wall -pedantic-errors -Wextra -Wno-long-long -Wno-deprecated-declarations -Wno-unused-parameter -Wno-variadic-macros -pthread"
+      ;;
+    Darwin)
+      CASSANDRA_CFLAGS="-Wall -pedantic -Wextra -Wno-long-long -Wno-deprecated-declarations -Wno-unused-parameter -Wno-variadic-macros"
+      ;;
+  esac
+  CASSANDRA_LIBS="-lssl -lz -luv -lstdc++"
```
avalanche123: why does the driver depend on stdc++ ?

mikefero: This was taken from the install.sh and Vagrantfile scripts where the LIBS were supplied, but it should be for the cpp-driver.

mikefero: This is where some cleanup should be done and some additional library checks as well since we are relying on things for the php-driver.

avalanche123: Right, I agree that we need to add checks for libuv since it is used in the driver, but we don't use any of the stdc++ things directly in the driver, that's why those flags were supplied from the outside for static linking the cpp-driver only

mikefero:
strndup is still used in php_cassandra.c (https://github.com/datastax/php-driver/blob/PHP-1/ext/php_cassandra.c#L656) which is why I added it back in ... will update to use estrndup.

ext/config.m4
mikefero: Reverted the changes back to the way they were. I still feel that all these build settings should be in the config.m4 and not be required outside the build like it is currently.

avalanche123:
Ah I thought I've removed it already, definitely feel free to change that

PR #25, opened 2015-05-04
PHP-16 - Numerics

mpenick:
Still working on tests. Decimal::sqrt() not implemented.


mpenick:
Additional: What are your thoughts on two subclasses to Numeric: Integer and FloatingPoint?

Current interfaces would like this:

Integer:
  Integer mod(Integer);

FloatPoint:
  FloatingPoint ceil();
  FloatingPoint floor();

avalanche123:
I think just including those methods everywhere and making them raise a LogicException when they are not applicable is what I'd prefer

avalanche123:
sorry, closed by mistake :)

PR #26, opened 2015-05-05
PHP-16 

mpenick:


ext/src/Cassandra/Exception/DivideByZeroException.c
```diff
@@ -0,0 +1,15 @@
...
+static zend_function_entry DivideByZeroException_methods[] = {
+  PHP_FE_END
+};
+
+void cassandra_define_DivideByZeroException(TSRMLS_D)
+{
+  zend_class_entry ce;
+
+  INIT_CLASS_ENTRY(ce, "Cassandra\\Exception\\DivideByZeroException", DivideByZeroException_methods);
+  cassandra_divide_by_zero_exception_ce = zend_register_internal_class_ex(&ce, cassandra_configuration_exception_ce, NULL TSRMLS_CC);
```
avalanche123: do you mind if we make this a subclass of `Cassandra\RangeException`?

ext/src/Cassandra/Types/Float.c
```diff
@@ -0,0 +1,446 @@
...
+{
+  cassandra_float* self =
+      (cassandra_float*) zend_object_store_get_object(getThis() TSRMLS_CC);
+
+  RETURN_DOUBLE((double) self->value);
+}
+/* }}} */
+
+/* {{{ Cassandra\Types\Float::minValue() */
+PHP_METHOD(Float, minValue)
```
avalanche123: I like these methods, I've seed similar ones in other classes, what do you think about making the static? E.g.:

```php
<?php
$maxFloat = Cassandra\Float::max();
```

ext/src/Cassandra/Types/Float.c
```diff
@@ -0,0 +1,446 @@
...
+{
+  object_init_ex(return_value, cassandra_float_ce);
+  cassandra_float* flt =
+          (cassandra_float*) zend_object_store_get_object(return_value TSRMLS_CC);
+  flt->value = LONG_LONG_MIN;
```
avalanche123: Looks like this one is using a long instead of double

ext/src/Cassandra/Types/Float.c
avalanche123: Heh, never mind it is static

ext/src/Cassandra/Types/Float.c
```diff
@@ -0,0 +1,446 @@
...
+ZEND_END_ARG_INFO()
+
+static zend_function_entry cassandra_float_methods[] = {
+  PHP_ME(Float, __construct, arginfo__construct, ZEND_ACC_CTOR|ZEND_ACC_PUBLIC)
+  PHP_ME(Float, __toString, arginfo_none, ZEND_ACC_PUBLIC)
+  PHP_ME(Float, value, arginfo_none, ZEND_ACC_PUBLIC)
+  PHP_ME(Float, isInfinite, arginfo_none, ZEND_ACC_PUBLIC)
+  PHP_ME(Float, isFinite, arginfo_none, ZEND_ACC_PUBLIC)
+  PHP_ME(Float, isNaN, arginfo_none, ZEND_ACC_PUBLIC)
+  PHP_ME(Float, add, arginfo_num, ZEND_ACC_PUBLIC)
```
avalanche123: would be cool to typehint these methods to accept `Cassandra\Numeric` instances

ext/src/Cassandra/Types/Float.c
mpenick: woah, what's going on there. fixing.

ext/src/Cassandra/Types/Float.c
mpenick: I thought about that, but I was thinking these could also accept php long/double/string?

ext/src/Cassandra/Exception/DivideByZeroException.c
mpenick: Makes sense. Changed.

ext/src/Cassandra/Types/Float.c
avalanche123: good point, we might relax the constraint later as that wouldn't be a BC break, but I think it makes sense to begin with a stricter interface

ext/src/Cassandra/Numeric.c
```diff
@@ -0,0 +1,35 @@
...
+  PHP_ABSTRACT_ME(Numeric, add, arginfo_num)
+  PHP_ABSTRACT_ME(Numeric, sub, arginfo_num)
+  PHP_ABSTRACT_ME(Numeric, mul, arginfo_num)
+  PHP_ABSTRACT_ME(Numeric, div, arginfo_num)
+  PHP_ABSTRACT_ME(Numeric, mod, arginfo_num)
+  PHP_ABSTRACT_ME(Numeric, abs, arginfo_none)
+  PHP_ABSTRACT_ME(Numeric, neg, arginfo_none)
+  PHP_ABSTRACT_ME(Numeric, sqrt, arginfo_none)
+  PHP_ABSTRACT_ME(Numeric, toInt, arginfo_none)
+  PHP_ABSTRACT_ME(Numeric, toDouble, arginfo_none)
```
avalanche123, line 23: This should be called `toFloat` to match PHP's type name. Is there a reason you didn't include `toVarint` `toDecimal` here?

mpenick, line 23: `toFloat()` is pretty confusing with there being a `Cassandra\Float` type. There are also "double" aliases for most "float" related things in PHP (unlike for "int"). It came down to time for `toVarint` etc. I didn't implement them so I didn't add them.

avalanche123, line 23: Got it, makes sense

PR #30, opened 2015-05-18
Feature regression tests

avalanche123:


Vagrantfile
```diff
@@ -42,7 +42,7 @@ Vagrant.configure("2") do |config|
   echo "Compiling and installing the extension..."
   pushd /usr/local/src/php-driver/ext
   phpize
-  LDFLAGS="-L$builddir/lib" LIBS="-lssl -lz -luv -lgmp -lstdc++" ./configure --with-cassandra=/tmp/php-driver-installation/build --with-gmp=/tmp/php-driver-installation/build --with-libdir=lib
+  LDFLAGS="-L$builddir/lib" LIBS="-lssl -lz -luv -lm -lgmp -lstdc++" ./configure --with-cassandra=/tmp/php-driver-installation/build --with-gmp=/tmp/php-driver-installation/build --with-libdir=lib
```
avalanche123, line 45: why do we depend on -lm?

support/ccm.php
```diff
@@ -94,16 +94,22 @@ public function stop()
...
     }
 
-    public function setup()
+    private function getClusters()
     {
         $clusters = array();
-        foreach (explode("\n", $this->run('list')) as $cluster) {
-            if ($this->isActive($cluster)) {
-                $active = count($clusters);
+        foreach (explode("\r\n", $this->run('list')) as $cluster) {
```
avalanche123: let's use `PHP_EOL` instead of `"\r\n"` here.

features/bootstrap/FeatureContext.php
```diff
@@ -155,6 +157,41 @@ public function itsOutputShouldContain(PyStringNode $string)
         PHPUnit_Framework_Assert::assertContains((string) $string, $this->getOutput());
     }
 
+    /**
+     * @Given the following logger settings:
+     */
+    public function theFollowingLoggerSettings(PyStringNode $string)
+    {
+        $lines = preg_split("/\\r\\n|\\r|\\n/", $string->getRaw());
```
avalanche123, line 315: you can use `explode(PHP_EOL, $string->getRaw())` here

Vagrantfile
mikefero, line 45: Logarithmic functions are being used (https://github.com/datastax/php-driver/blob/master/ext/util/math.c#L427) and were not being linked in (e.g. undefined references)

support/ccm.php
mikefero: Good catch

features/bootstrap/FeatureContext.php
mikefero, line 315: Tried that and the result was one line returned from explode when on Windows platform. I can continue to play around with other options.

Vagrantfile
avalanche123, line 45: gotcha, thanks!

features/bootstrap/FeatureContext.php
avalanche123, line 315: nah, leave whatever works in

PR #35, opened 2015-06-04
New remove method for Collection

SachaMorard:
Hi
I was migrating my app from thrift to your new extension, and I had a problem to remove value of a LIST type.
Do you think it would be usefull ?

I'm not good at all in C language, so don't hesitate to check twice ;)


avalanche123:
merged and fixed wrong type info, thanks!

PR #37, opened 2015-07-02
ext/doc fixes

r3nat:
http://www.phpdoc.org/docs/latest/guides/types.html#class-name says, that class names without slashes are relative


avalanche123:
does this actually prevent ide autocompletion for you? Because we are not using phpdocumentor for the api docs generation

r3nat:
yep

image

avalanche123:
kk, I'll merge this in after I confirm that generated docs still look good, thanks!

avalanche123:
merged and pushed, thanks!

PR #49, opened 2015-09-02
PHP-22

avalanche123:
- Schema Metadata
- Types API, e.g. `Cassandra\Type::map(Cassandra\Type::varchar(), Cassandra\Type::boolean())->create("does it work?", true);`
- updated CHANGELOG and package.xml


ext/doc/Cassandra/Cluster.php
```diff
@@ -30,7 +30,7 @@
      *
      * @return Session Session instance
      */
-    public function connect($keyspace = null);
+    function connect($keyspace = null);
```
jippi, line 33: normal php convention is to include the visibility and not rely on implicit :)

avalanche123, line 33: Not for interfaces, non public visibility makes little sense there

ext/config.w32
```diff
@@ -67,7 +67,22 @@ if (PHP_CASSANDRA != "no" &&
             "Timeuuid.c " +
             "Uuid.c " +
             "UuidInterface.c " +
-            "Varint.c", "cassandra");
+            "Varint.c" +
```
mikefero, line 70: Recommend merging this before #48 due to differences that will likely cause a merge conflict; sweeping file changes in the PR for the Windows PECL build.

ext/doc/Cassandra/DefaultKeyspace.php
```diff
@@ -0,0 +1,68 @@
...
+     *
+     * @return string Whether durable writes are enabled
+     */
+    public function hasDurableWrites() {}
+
+    /**
+     * {@inheritDoc}
+     *
+     * @param  string                $name  Table name
+     * @return Cassandra\Table|null         Table instance of null
```
mikefero: of = or

ext/doc/Cassandra/Keyspace.php
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Returns whether the keyspace has durable writes enabled
+     * @return string Whether durable writes are enabled
+     */
+    function hasDurableWrites();
+
+    /**
+     * Returns a table by name
+     * @param  string               $name Table name
+     * @return Cassandra\Table|null       Table instance of null
```
mikefero: same as DefaultKeyspace; of = or

ext/package.xml
```diff
@@ -14,36 +14,19 @@ protocol and Cassandra Query Language v3.
...
+    <api>1.0.0</api>
   </version>
   <stability>
-    <release>beta</release>
+    <release>stable</release>
     <api>stable</api>
   </stability>
   <license uri="http://www.apache.org/licenses/LICENSE-2.0">Apache License 2.0</license>
   <notes>
-# 1.0.0-rc
```
mikefero, line 29: Is the GA changelog being left off intentionally?

ext/src/Cassandra/DefaultColumn.c
```diff
@@ -0,0 +1,207 @@
...
+#include "util/result.h"
+#include "util/ref.h"
+
+zend_class_entry *cassandra_default_column_ce = NULL;
+
+PHP_METHOD(DefaultColumn, name)
+{
+  cassandra_column* self;
+
+  if (zend_parse_parameters_none() == FAILURE)
```
mikefero: Suggestion; single line if or use braces

ext/src/Cassandra/DefaultColumn.c
```diff
@@ -0,0 +1,207 @@
...
+  PHP_ME(DefaultColumn, indexOptions, arginfo_none, ZEND_ACC_PUBLIC)
+  PHP_FE_END
+};
+
+static zend_object_handlers cassandra_default_column_handlers;
+
+static HashTable*
+php_cassandra_default_column_properties(zval *object TSRMLS_DC)
+{
+  /* cassandra_column* self = (cassandra_column*) zend_object_store_get_object(object TSRMLS_CC); */
```
mikefero: leftover?

ext/src/Cassandra/DefaultSchema.c
```diff
@@ -0,0 +1,152 @@
...
+  PHP_ME(DefaultSchema, keyspaces, arginfo_none, ZEND_ACC_PUBLIC)
+  PHP_FE_END
+};
+
+static zend_object_handlers cassandra_default_schema_handlers;
+
+static HashTable*
+php_cassandra_default_schema_properties(zval *object TSRMLS_DC)
+{
+  /* cassandra_schema* self = (cassandra_schema*) zend_object_store_get_object(object TSRMLS_CC); */
```
mikefero: leftover?

ext/src/Cassandra/DefaultSchema.c
```diff
@@ -0,0 +1,152 @@
...
+{
+  cassandra_schema* schema = (cassandra_schema*) object;
+
+  zend_object_std_dtor(&schema->zval TSRMLS_CC);
+
+  if (schema->schema) {
+    php_cassandra_del_ref(&schema->schema);
+  }
+
+  efree(schema);
```
mikefero, line 115: Is it possible to still have references to the schema before the freeing occurs?

https://github.com/datastax/php-driver/pull/49/files#diff-42ddcbab12a61279ba8ac85265a2d2c7R63

ext/src/Cassandra/DefaultTable.c
```diff
@@ -0,0 +1,411 @@
...
+
+  if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "s", &name, &name_len) == FAILURE) {
+    return;
+  }
+
+  self     = (cassandra_table*) zend_object_store_get_object(getThis() TSRMLS_CC);
+  metadata = cass_schema_meta_get_entry(self->meta, name);
+
+  if (metadata == NULL) {
+    RETURN_NULL();
```
mikefero: nitpick: Inconsistency for NULL value return; unless RETURN_NULL() is doing something different than return null.

ext/src/Cassandra/Type.c
```diff
@@ -0,0 +1,195 @@
...
+                           "Cassandra\\Type::inet(), " \
+                           "Cassandra\\Type::timestamp(), " \
+                           "Cassandra\\Type::uuid() or " \
+                           "Cassandra\\Type::timeuuid()" TSRMLS_CC);
+    return;
+  }
+
+  if (!instanceof_function(Z_OBJCE_P(value_type),
+                           cassandra_type_scalar_ce TSRMLS_CC)) {
+    throw_invalid_argument(value_type, "valueType",
```
mikefero, line 131: A helper function might help reduce the amount of duplication.

avalanche123, line 131: good call

ext/util/types.c
```diff
@@ -0,0 +1,657 @@
+#include "php_cassandra.h"
+#include "util/types.h"
+#include <ext/standard/php_smart_str.h>
+#include "../src/Cassandra/Bigint.h"
```
mikefero: You shouldn't need the relative parent paths here; root include search path should be the **ext** directory. In addition it would be good to add **src/Cassandra** to the include path in the config.m4 (already exists for Windows config.w32). With that the paths could be removed from the include statement.

Since header guards are used for all the Cassandra/*.h there will not be issues with system headers (e.g. float.h, math.h, ...etc). I would even suggest just include the **src** directory in the include search path and then the include line could be `#include "Cassandra/Float.h"`; as to not confuse anybody reading through the code and mistaken `Float.h` for `float.h` (which is more of an issues for case-insensitive operating systems; WIndows for example)

avalanche123: will do

ext/util/types.c
```diff
@@ -0,0 +1,657 @@
...
+};
+
+static const char*
+describe_token(enum token_type token)
+{
+  switch (token) {
+  case TOKEN_ILLEGAL:     return "illegal character";      break;
+  case TOKEN_PAREN_OPEN:  return "opening parenthesis";    break;
+  case TOKEN_PAREN_CLOSE: return "closing parenthesis";    break;
+  case TOKEN_COMA:        return "coma";                   break;
```
mikefero: `TOKEN_COMMA` and `return "comma";`

ext/util/types.c
```diff
@@ -0,0 +1,657 @@
...
+      case '(':
+        type = TOKEN_PAREN_OPEN;
+        i++;
+        break;
+      case ')':
+        type = TOKEN_PAREN_CLOSE;
+        i++;
+        break;
+      case ',':
+        type = TOKEN_COMA;
```
mikefero: `TOKEN_COMMA`

ext/util/types.c
```diff
@@ -0,0 +1,657 @@
...
+  return type;
+}
+
+struct node_s {
+  struct node_s* parent;
+  const char*    name;
+  size_t         name_length;
+  struct node_s* first_child;
+  struct node_s* last_child;
+  struct node_s* next_sibbling;
```
mikefero: misspelling: **sibbling** to sibling

ext/util/types.c
```diff
@@ -0,0 +1,657 @@
...
+        if (node->parent == NULL) {
+          EXPECTING_TOKEN("end of string");
+        }
+
+        node = node->parent;
+        continue;
+      } else if (token == TOKEN_END) {
+        break;
+      } else {
+        EXPECTING_TOKEN("a coma, a closing parenthesis or an end of string");
```
mikefero: comma

ext/util/types.c
```diff
@@ -0,0 +1,657 @@
...
+        node = child;
+        continue;
+      } else if (token == TOKEN_PAREN_CLOSE) {
+        state = STATE_AFTER_PARENS;
+        continue;
+      } else if (token == TOKEN_END) {
+        break;
+      } else {
+        php_cassandra_parse_node_free(root);
+        EXPECTING_TOKEN("opening/closing parenthesis or coma");
```
mikefero: comma

ext/src/Cassandra/DefaultSchema.c
avalanche123, line 115: yes, something like the following:

``` php
<?php

$keyspace = $session->schema()->keyspace("abc");
```

the Cassandra\Keyspace object would have a reference to CassSchema, while the Cassandra\Schema object would get GC'd by falling out of scope, since we don't preserve the result of `schema()` call anywhere.

Did I understand your question correctly?

mikefero, line 115: OK that makes sense ... thanks for clearing that up.

mikefero:
Sounds good ... :shipit:

avalanche123:
merged!

PR #50, opened 2015-09-04
Avoid error when getting timeout or arguments if value is null in ExecutionOption

mojoLyon:
``` php
$options = new \Cassandra\ExecutionOptions(array('arguments' => $song));
var_dump($options->arguments); // ok

$options = new \Cassandra\ExecutionOptions(array('timeout' => 5));
var_dump($options->arguments); // ko core dump
```

```
0  zim_ExecutionOptions___get (ht=<optimized out>, return_value=0x7f7d4a79ebf8, return_value_ptr=<optimized out>, this_ptr=<optimized out>, return_value_used=<optimized out>)
    at /home/stephane/php-driver/ext/src/Cassandra/ExecutionOptions.c:93
93      RETURN_ZVAL(self->arguments, 1, 0);
```


avalanche123:
Thanks, updated and merged!

PR #53, opened 2015-11-17
fix segfault on rows->nextPageAsync

Developerius:


avalanche123:
looks good, thanks for the fix!

PR #56, opened 2015-11-26
Create config.m4

madz0:
When compiling for the installed cpp-driver, one might get configure: error: Unable to load libcassandra error. In config.log detailed error says undefined reference to `deflateInit_' so I added -lz to fix this

### DIFF
--- a/ext/config.m4
+++ b/ext/config.m4
@@ -217,6 +217,10 @@ if test "$PHP_CASSANDRA" != "no"; then
     done
   fi
 
+#changed by mohamad zeynali mohammad.basu@gmail.com (apache license) to fix undefined reference to `deflateInit_' bug
+##
+LIBS="-lz $LIBS"
+
   if test -z "$CPP_DRIVER_DIR"; then
     ac_extra=
   else
--- a/ext/config.m4
+++ b/ext/config.m4
@@ -218,7 +218,7 @@ if test "$PHP_CASSANDRA" != "no"; then
   fi
 
 #changed by mohamad zeynali mohammad.basu@gmail.com (apache license) to fix undefined reference to `deflateInit_' bug
-##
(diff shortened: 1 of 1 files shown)


ext/config.m4
```diff
@@ -217,6 +217,10 @@ if test "$PHP_CASSANDRA" != "no"; then
     done
   fi
 
+#changed by mohamad zeynali mohammad.basu@gmail.com (apache license) to fix undefined reference to `deflateInit_' bug
```
mpenick, line 220: Thanks for the pull request. Please move this comment into the commit message.

madz0, line 220: Thanks for the regards. Would you like me to amend df8d7b2 commit or just create another commit with the mentioned comment? I just created a comment on the commit.

mpenick, line 220: Amend would be preferred, but either way is fine.

madz0, line 220: My previous commit was changed due to some auto generated build files. I
committed a new one with the provided commit message. Hope this does the
job here is the commit id 842915e04e521d9e62edda7187d7098ffdcefdf7
and this is the direct link
https://github.com/mohamad-z/php-driver/commit/842915e04e521d9e62edda7187d7098ffdcefdf7

mpenick:
Not all builds require "-lz". Could you add this as an optional library to config.m4 (without the comment please)? Otherwise, you can do this via:

``` sh
LDFLAGS='-lz' ./configure
```

mpenick:
Let me know if you're going make it an optional library or please close the pull request.

madz0:
Ok, I see. Unfortunately I'm so busy for couple of days. If you prefer then
I close the pull for now, but let me know if you wanted this to be
implemented and next time no comment  ;)

PR #58, opened 2015-12-21
62 - PHP7 and Nested collections

mpenick:


ext/config.m4
```diff
@@ -89,6 +89,7 @@ if test "$PHP_CASSANDRA" != "no"; then
     src/Cassandra/Set.c \
     src/Cassandra/Map.c \
     src/Cassandra/Collection.c \
+    src/Cassandra/Value.c \
```
mikefero, line 92: This should be added to `ext/config.w32` for Windows builds too.

ext/doc/Cassandra.php
```diff
@@ -322,7 +322,12 @@
     /**
      * Current version of the extension.
      */
-    const VERSION = '1.0.0-beta';
+    const VERSION = '1.1';
+
+    /**
+     * Version of the cpp-driver the extension is compiled against.
+     */
+    const CPP_DRIVER_VERSION = '2.2.2';
```
mikefero, line 330: This might not necessarily be true in the event a different version of the cpp-driver is used. Can we use the [defines in `cassandra.h`](https://github.com/datastax/cpp-driver/blob/master/include/cassandra.h#L53-L56) instead to be dynamic?

ext/php_cassandra.c
```diff
@@ -388,85 +391,16 @@ PHP_RINIT_FUNCTION(cassandra)
...
-  if (CASSANDRA_G(type_timestamp)) {
-    zval_ptr_dtor(&CASSANDRA_G(type_timestamp));
-    CASSANDRA_G(type_timestamp) = NULL;
-  }
-
-  if (CASSANDRA_G(type_uuid)) {
-    zval_ptr_dtor(&CASSANDRA_G(type_uuid));
-    CASSANDRA_G(type_uuid) = NULL;
-  }
+#define TYPE_CODE(m) type_ ## m
```
mikefero: Very nice :exclamation:

ext/doc/Cassandra.php
mikefero, line 330: Nevermind ... This looks like it is just for the documentation stubs and not what is [built into the library](https://github.com/datastax/php-driver/pull/58/files#diff-d3776f8e62027b140044da1533b29c4fR82)

ext/doc/Cassandra/Bigint.php
```diff
@@ -45,6 +45,13 @@ public static function max() {}
     public function __construct($value) {}
 
     /**
+     * The type of this bigint.
+     *
+     * @return Type
+     */
+    public function type() {}
```
avalanche123, line 52: missing `implements Value` ?

ext/doc/Cassandra/Decimal.php
```diff
@@ -41,6 +41,13 @@
     public function __construct($value) {}
 
     /**
+     * The type of this decimal.
+     *
+     * @return Type
+     */
+    public function type() {}
```
avalanche123, line 48: missing `implements Value` ?

ext/doc/Cassandra/Float.php
```diff
@@ -45,6 +45,13 @@ public static function max() {}
     public function __construct($value) {}
 
     /**
+     * The type of this float.
+     *
+     * @return Type
+     */
+    public function type() {}
```
avalanche123, line 52: missing `implements Value` ?

ext/doc/Cassandra/Numeric.php
```diff
@@ -27,7 +27,7 @@
  * @see Float
  * @see Varint
  */
-interface Numeric
+interface Numeric implements Value
```
avalanche123: should be `extends` here, since `Numeric` is also an interface, then disregard previous comments

ext/doc/Cassandra/Float.php
avalanche123, line 52: ignore, it is `Numeric`

ext/doc/Cassandra/Decimal.php
avalanche123, line 48: ignore, it is `Numeric`

ext/doc/Cassandra/UuidInterface.php
```diff
@@ -21,7 +21,7 @@
 /**
  * A PHP representation of the CQL `timeuuid` datatype
  */
-interface UuidInterface
+interface UuidInterface implements Value
```
avalanche123: also `extends`

avalanche123:
The changes look good minus two things I've pointed out.
As a design suggestion, I'd prefer to keep `Numeric` and `UuidInterface` interfaces unaware of the `Value` interface and implement all interfaces in concrete classes themselves because those interfaces are unrelated.

ext/src/Cassandra/Timeuuid.c
```diff
@@ -147,12 +157,12 @@ php_cassandra_timeuuid_properties(zval *object TSRMLS_DC)
 
   char* string = emalloc((CASS_UUID_STRING_LENGTH) * sizeof(char));
```
mikefero: Is `string` or `self->uuid` freed when calling `efree(self)` in [`php_cassandra_timeuuid_free`](https://github.com/datastax/php-driver/pull/58/files#diff-48d0a9b7bb2a2633ad51f5b3e165e6ebR207)? Just checking in case we have a memory leak.

ext/util/collections.c
```diff
@@ -15,160 +17,507 @@
...
-  }
 
-  switch (type) {
-  case CASS_VALUE_TYPE_ASCII:
+  cassandra_type* type = (cassandra_type*) zend_object_store_get_object(ztype TSRMLS_CC);
+
+  switch (type->type) {
   case CASS_VALUE_TYPE_VARCHAR:
   case CASS_VALUE_TYPE_TEXT:
+  case CASS_VALUE_TYPE_ASCII:
```
mikefero, line 33: Good catch.

ext/src/Cassandra/Collection.c
```diff
@@ -42,18 +46,20 @@ php_cassandra_collection_find(cassandra_collection* collection, zval* object, lo
...
   HashPointer ptr;
   zval**      current;
-  zval        compare;
   ulong       idx;
+  cassandra_type_collection* type;
+
+  type = (cassandra_type_collection*) zend_object_store_get_object(collection->type TSRMLS_CC);
 
-  if (!php_cassandra_validate_object(object, collection->type TSRMLS_CC))
+  if (!php_cassandra_validate_object(object, type->value_type TSRMLS_CC))
```
mikefero: Since the return is on the next line how do you feel about bracing or inlining the return?

ext/src/Cassandra/Collection.c
```diff
@@ -91,36 +97,53 @@ php_cassandra_collection_populate(cassandra_collection* collection, zval* array)
...
 PHP_METHOD(Collection, __construct)
 {
-  char *type;
-  int type_len;
-  cassandra_collection* collection;
+  cassandra_collection* self;
+  zval* value_type;
 
-  if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "s", &type, &type_len) == FAILURE) {
+  if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "z", &value_type) == FAILURE)
```
mikefero, line 77: Same here.

ext/src/Cassandra/Map.c
```diff
@@ -28,169 +24,227 @@ php_cassandra_map_set(cassandra_map* map, zval* zkey, zval* zvalue TSRMLS_DC)
...
-  char* key_type;
-  char* value_type;
-  int key_type_len, value_type_len;
-  cassandra_map* map;
+  cassandra_map* self;
+  zval* key_type = NULL;
+  zval* value_type = NULL;
 
-  if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "ss", &key_type, &key_type_len, &value_type, &value_type_len) == FAILURE) {
+  if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "zz", &key_type, &value_type) == FAILURE)
```
mikefero, line 161: and here.

ext/src/Cassandra/Set.c
```diff
@@ -7,130 +9,162 @@ zend_class_entry *cassandra_set_ce = NULL;
...
 PHP_METHOD(Set, __construct)
 {
-  char *type;
-  int type_len;
-  cassandra_set* set;
+  cassandra_set* self;
+  zval* value_type;
 
-  if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "s", &type, &type_len) == FAILURE) {
+  if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "z", &value_type) == FAILURE)
```
mikefero, line 105: and here.

mikefero:
LGTM (pending #59 merge); the updated memory management and code consistency (e,g, using self and type consistently) is an added bonus.

There are a bunch more `if` blocks without bracing or inlined; I was just trying to go for consistency here as that appears to be the predominant style used.

jippi:
Nice with PHP7 support - will this also add support for C\* 3.x ?

emulienfou:
Great job, just notice 2 problems with the **install.sh** script:
1. Got `configure: error: Unable to load libcassandra`, due to this line: `rm -f build/lib/libcassandra.{dylib,so}`
2. Got `PHP Warning:  PHP Startup: Unable to load dynamic library '/usr/lib64/php/modules/cassandra.so' - libcassandra.so.2: cannot open shared object file: No such file or directory in Unknown on line 0` error, due to this line: `rm -Rf /tmp/php-driver-installation`. For this error I needed to copy **libcassandra.so.2** from `/tmp/php-driver-installation/lib` folder into `/usr/lib/libcassandra.so.2`

I think there is a problem with the install script. Not expert with that however !!!

mpenick:
@avalanche123 Thanks for reviewing. Good idea!

To be clear you're advocating this?

``` php
interface Numeric {
    public function value();
}

interface Value {
    public function type();
}

class Bigint implements Value, Numeric {
    public function type() {
    }
    public function value() {
    }
}
```

mpenick:
@jippi When using cpp-driver > 2.2.2 you should be able to to connect to connect to and use C\* 3.0 but you won't be able to use some C\* 2.2/3.0 features (e.g. tinyint/smallint types, date/time types, UNSET, new schema metadata).

jippi:
@mpenick okay, so 1.0.1 doesn't support 2.2.2 ? its pinned to 2.2.1 in the submodule from what I can see :)

avalanche123:
@mpenick indeed, your understanding is correct

PR #59, opened 2016-01-04
39 - PHP7

mpenick:


mpenick:
PHP7 is now here: https://github.com/datastax/php-driver/pull/58

PR #67, opened 2016-01-20
60 - Support for retry policies

mpenick:


jippi:
its quite confusing that all PR include the commits for 2.2.x support and back :(

jippi:
can you stack them either as a chain of attempts to pop, or to run both `Retry` + `Log`? 

Or try a chain of policies like `Retry`, `Downgrade`, `Fallthrough`  ?

mpenick:
@jippi Unfortunately, PHP7 required a bunch of changes. It's not fun retroactively adding PHP7 support to a PR. We did that with nested collections and the merge was time consuming and error prone. Now that that merge is done we're treating https://github.com/datastax/php-driver/tree/62 as our "master" for this release. This is mostly because reviews haven't kept pace with the features being added.  In the future I'll try to merge to actual "master" more frequently.

mpenick:
@jippi Just merged 62 into master. That should make the PR cleaner. I rebased on master and force pushed so if you're using this you'll need to update your local branch.

jippi:
Thanks! :) 
Any thoughts on my idea for chained/stacked retry policies?

mpenick:
@jippi I don't fully understand what you mean by chained/stacked retry polices. Could you explain more about what you would want to achieve with that?

Currently there are 3 categories of policies:
- Retry when it when it won't affect consistency (The default policy)
- Retry whenever it will make the request succeed even if it affects consistency (The downgrade policy)
- Always fail (The fall-through policy)

Logging can be combined with any of the above. 

This currently represents what the underlying C/C++ driver supports.  It is possible that in the future we may allow PHP implementations of these policies; however, this may be very difficult as PHP has limited support for running interpreted code in multiple threads.

PR #71, opened 2016-03-07
Blacklist dc

sandro-lex:
Add support to black/white list policies


ext/src/Cassandra/Cluster/Builder.c
```diff
@@ -755,6 +968,14 @@ static zend_function_entry cassandra_cluster_builder_methods[] = {
          ZEND_ACC_PUBLIC)
   PHP_ME(ClusterBuilder, withDatacenterAwareRoundRobinLoadBalancingPolicy,
          arginfo_dc_aware, ZEND_ACC_PUBLIC)
+  PHP_ME(ClusterBuilder, withBlackListNodes,
```
mpenick: What do you think about "withBlackListHosts" and "withWhiteListHosts" instead of "Nodes"?

ext/src/Cassandra/Cluster/Builder.c
```diff
@@ -755,6 +968,14 @@ static zend_function_entry cassandra_cluster_builder_methods[] = {
...
+         arginfo_blacklist_nodes, ZEND_ACC_PUBLIC)
+  PHP_ME(ClusterBuilder, withWhiteListNodes,
+          arginfo_whitelist_nodes, ZEND_ACC_PUBLIC)
+  PHP_ME(ClusterBuilder, withBlackListDC,
```
mpenick: "DCs"? (plural)

mpenick:
Thanks for this!

craveica:
I think the reference for cpp-driver should be updated.

```
php -n -dextension='.libs/cassandra.so' -m
PHP Warning:  PHP Startup: Unable to load dynamic library '.libs/cassandra.so' - .libs/cassandra.so: undefined symbol: cass_cluster_set_whitelist_dc_filtering in Unknown on line 0
```

PR #76, opened 2016-04-05
disable obj cloning and check for NULL before freeing

tony2001:
The patch fixes segfault in the following code:

``` php
$Class = new ReflectionClass('Cassandra\FutureSession');
var_dump($Class->isCloneable());
```

and also disables object cloning, as it requires special clone handlers in order to clone internal object structures.


mpenick:
Thanks!

PR #77, opened 2016-04-07
63 and 64: tinyint/smallint and time/data types

avalanche123:
This is a work in progress.
Unfortunately, I think I've uncovered a bug in metadata parsing in cpp driver.
The attached test case always causes "Invalid value" exception.
After adding a print statement in `execute_request.hpp`, `get_type()`, I see that the type of value is considered a custom type `org.apache.cassandra.db.marshal.ShortType` instead of `smallint`.

@mpenick let me know if you need help reproducing or if it turns out that my implementation is at fault


mpenick:
what cpp-driver function is returning the error? Is it binding a value to a prepared statement? I've tested this using C\* 2.2 and 3.X and haven't been able to reproduce it yet.

mpenick:
Here's the example used: https://gist.github.com/mpenick/c70e743cbed922d9e5d3c587874c82da

ext/src/Cassandra/Date.c
```diff
@@ -97,14 +100,51 @@ PHP_METHOD(Date, toDateTime)
...
+  }
+
+  if (!datetime_obj->time) {
+    zend_throw_exception_ex(cassandra_invalid_argument_exception_ce, 0 TSRMLS_CC,
+                            "DateTime object has not been correctly initialized");
+    return;
+  }
+  timelib_update_ts(datetime_obj->time, NULL);
+
+  timestamp = timelib_date_to_int(dateobj->time, &error);
```
mikefero: **`timelib_update_ts`** and **`timelib_date_to_int`** is an internal function and is not exported in the PHP dynamic library. Linking the extension statically into the executable is possible; however a DLL is not.  Since these functions are used in both `Time.c` and `Date.c` a utils implementation would be best.

jippi:
Any alive-ness of this one? 

Also, are there a way to insert `date` fields without this PR ?

mpenick:
@jippi sorry, been working on the cpp-driver. We're working on getting this out ASAP.

jippi:
Thanks - cost a lot of grief last week :(

mpenick:
Merged at 0d2159ee8dfe2feaa50c5ccf164f032a9b990874

PR #81, opened 2016-06-06
Fix PHP documentation for Type/Collection and Type/Set

geofflee:
Corrected type() method name to valueType() for Type/Collection and Type/Set in the PHP documentation.


mpenick:
Thanks for the corrections.

PR #86, opened 2016-08-01
Session Performance/Diagnostic Metrics

mikefero:


mpenick:
Looks good. Simple and clean.

avalanche123:
Just as an aside, I think it would be beneficial for the users of our API to define metrics object/interface with methods such as `min()`, `max()`, etc. this would make it harder for us to break backwards compatibility later and will also make it easier to document what those do. Additionally, interfaces can be mocked in tests, and objects with an API can be autocompleted in IDEs. So good things all around :)

mpenick:
@avalanche123 @mikefero Agreed. That would be better here because the underlying data is concrete. I think this is appealing because it's soooo much less boilerplate, but it makes sense in this case to make it an interface/class.

engharb:
Dears,
it will be kind of you to help me. I am a gegginer in How to integrate such API in my PHP projects. and I want to use the Cassandra PHP-Driver (I configured the Java-Driver and using it successfully). I tried to follow the stps described but it does not work on Ubuntu. Is there a way to install it via composer ? if yes How?
Or give another easier way to install the Driver?

Best regards

mikefero:
@engharb Please use the [mailing list](https://groups.google.com/a/lists.datastax.com/forum/#!forum/php-driver-user) or IRC channel (#datastax-drivers on irc.freenode.net) in order to help you with any questions (like these) that you have concerning the PHP driver.

Thank you.

PR #91, opened 2017-02-06
PHP-143 - Memory leak in `FutureSession`

mpenick:


mpenick:
Reproduced using:

```php
<?php

$cluster = Cassandra::cluster()->build();
$connectionFuture = $cluster->connectAsync();
$session = $connectionFuture->get();
```

mikefero (approved):
+1

PR #92, opened 2017-02-07
PHP-86 - Decimal __toString creates incorrect string representation

stamhankar999:
* Also fixed defect where parsing a double resulted in a garbage Decimal object.

### DIFF
--- a/ext/util/math.c
+++ b/ext/util/math.c
@@ -424,33 +424,11 @@ php_driver_parse_decimal(char *in, int in_len, mpz_t *number, long *scale TSRMLS
 void
 php_driver_format_integer(mpz_t number, char **out, int *out_len)
 {
-  size_t len;
-  char *tmp;
-
-  len = mpz_sizeinbase(number, 10);
-  if (mpz_sgn(number) < 0)
-    len++;
-
-  tmp = (char*) emalloc((len + 1) * sizeof(char));
-  mpz_get_str(tmp, 10, number);
-
-  // mpz_sizeinbase returns the length of the number, but it can be 1 more than the actual length.
-  // For example, for value 950, it returns 4. We want an accurate length, so we check the len-1
-  // byte to see if it is the null terminator; if it is, our real length is actually one less.
-  //
-  // If len is accurate to start with, tmp[len-1] is the last digit. Force tmp[len] to be the null-terminator.
-  // sandman suggests: This seems unnecessary. The if condition assumes that mpz_get_str returns a null-terminated
-  // string, so the else clause shouldn't be paranoid about this.
-  if (tmp[len - 1] == '\0') {
-    len--;
-  } else {
-    tmp[len] = '\0';
-  }
-
-  *out     = tmp;
-  *out_len = len;
+  /* Adding 2 ensures enough space for the null-terminator and negative sign */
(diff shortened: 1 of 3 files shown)


ext/util/math.c
```diff
@@ -434,6 +434,13 @@ php_driver_format_integer(mpz_t number, char **out, int *out_len)
   tmp = (char*) emalloc((len + 1) * sizeof(char));
   mpz_get_str(tmp, 10, number);
 
+  // mpz_sizeinbase returns the length of the number, but it can be 1 more than the actual length.
```
mpenick: A potential simplification:
```c
void
php_driver_format_integer(mpz_t number, char **out, int *out_len)
{
  /* Adding 2 ensures enough space for the null-terminator and negative sign */
  *out = (char*) malloc(mpz_sizeinbase(number, 10) + 2);
  mpz_get_str(*out, 10, number);
  *out_len = strlen(*out);
}
```

Source: https://gmplib.org/manual/Converting-Integers.html

ext/util/math.c
```diff
@@ -462,91 +469,98 @@ php_driver_format_decimal(mpz_t number, long scale, char **out, int *out_len)
...
+  // Ultimately, we want to return a string representation of this decimal. So allocate
+  // a buffer that could hold this decimal in the worst possible conservative case.
+
+  // absolute length + negative sign + point sign + scale (in case we end up with a number with leading 0s) +
+  // exponent modifier and sign.
+  tmp = (char*) emalloc(len + negative + 1 + scale + 2);
+  mpz_get_str(tmp, 10, number);
+
+  // Update len to be the true length of the string representation of |number|.
+  if (tmp[len - 1] == '\0' || (tmp[len] == '\0' && negative)) {
```
mpenick: use `strlen()` to simplify?

ext/util/math.c
```diff
@@ -462,91 +469,98 @@ php_driver_format_decimal(mpz_t number, long scale, char **out, int *out_len)
...
+    len--;
+  }
+
   point = len - scale;
 
   if (scale >= 0 && (point - 1) >= -6) {
     if (point <= 0) {
-      /* current position */
+      // e.g. -0.002 and 0.002
+      int shift_start = negative ? 1 : 0;
```
mpenick: I think this can be: `int shift_start = negative`

ext/util/math.c
stamhankar999: Yeah, I saw that, too, but I thought our existing code was trying to avoid the strlen call.

ext/util/math.c
stamhankar999: doh! My head was so deep in it, I didn't see the obvious. Good catch!

ext/util/math.c
```diff
@@ -462,91 +447,95 @@ php_driver_format_decimal(mpz_t number, long scale, char **out, int *out_len)
   if (mpz_sgn(number) < 0)
     negative = 1;
 
+  // mpz_sizeinbase returns the length of |number|, but it can be off by one (e.g. 950 has length 4).
```
mpenick: > mpz_sizeinbase returns the length of |number|, but it can be off by one (e.g. 950 has length 4)

Does this need to be here anymore?

mpenick:
+1. Squash and merge.

PR #93, opened 2017-02-07
test: PHP-119 - Driver crash getting column type

mikefero:
This issue was isolated to PHP v5.6.x; however this test can be utilized to ensure correctness for all supported PHP versions.


mpenick:
My concern is the default schema can vary on different versions of C* so this might not test anything. I wonder if there some builtin test custom type we can use for this test?

mikefero:
True! Let me dig into more closely and get the exact problem type that was causing the issue in the first place.

mikefero:
Updated original test with a NULL check and created additional test for root cause; datastax/cpp-driver#345.

mpenick:
+1 squash and merge.

PR #94, opened 2017-02-07
PHP-139 - Support simple string queries

mpenick:
CQL strings should be able to be executed directly without the need to
be wrapped in a `Cassandra\SimpleStatement`.

Also added support to `Cassandra\BatchStatement::add()`.


mpenick:
Somewhat WIP. I haven't added a test for batch statements.

stamhankar999:
👍 Merge at will.

mpenick:
Merged at 1ca29c54b97d85548f4969e864c95ee076a659c0

PR #95, opened 2017-02-08
PHP-144 - Persistent sessions reference counting issue

mpenick:
Fixed issue where persistent sessions were incorrectly double decrementing the
reference count; Once when deleting from the hash and once in the persistent
session dtor.

### DIFF
--- a/ext/src/DefaultCluster.c
+++ b/ext/src/DefaultCluster.c
@@ -117,11 +117,7 @@ PHP_METHOD(DefaultCluster, connect)
 
   if (php_driver_future_is_error(future TSRMLS_CC) == FAILURE) {
     if (session->persist) {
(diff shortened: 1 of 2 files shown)


ext/src/DefaultCluster.c
```diff
@@ -117,11 +117,7 @@ PHP_METHOD(DefaultCluster, connect)
 
   if (php_driver_future_is_error(future TSRMLS_CC) == FAILURE) {
     if (session->persist) {
-      if (PHP5TO7_ZEND_HASH_DEL(&EG(persistent_list), hash_key, hash_key_len + 1)) {
```
jippi, line 120: `// done, no it was not :P`

mpenick: heh, that's exactly what I was thinking as I was adding this...

stamhankar999:
👍

mpenick:
Merged at f1300a21af7165305ab983a1e9c9c46be3d0ee6f

PR #97, opened 2017-02-08
PHP-86 - Decimal __toString broken

stamhankar999:


ext/src/Decimal.c
```diff
@@ -59,16 +59,18 @@ to_mpf(mpf_t result, php_driver_numeric *decimal)
...
 #define DOUBLE_EXPONENT_BITS 11
 #define DOUBLE_EXPONENT_MASK (cass_int64_t) ((1LL << DOUBLE_EXPONENT_BITS) - 1)
 
 static void
 from_double(php_driver_numeric *result, double value)
 {
-  cass_int64_t raw = *((cass_int64_t*) &value);
+  // Copy the bits of value into an int64 so that we can do bit manipulations on it.
+  cass_int64_t raw = 0;
+  memcpy(&raw, &value, 8);
```
mpenick: This can't be interleaved with the other variable declarations.

stamhankar999: Doh! The Windows build fails with such an arcane error; thanks a ton. I totally forgot about C89 (or whatever).

PR #98, opened 2017-02-09
test: PHP-143 - Memory leak in `FutureSession`

mikefero:


mikefero:
Closing PR as this test doesn't guarantee proper memory leak checking from the driver.

PR #99, opened 2017-02-09
test: PHP-127 - Adding duration type to data provider

mikefero:


tests/integration/Cassandra/DatatypeIntegrationTests.php
```diff
@@ -34,6 +34,9 @@ public function scalarCassandraTypes() {
             array(Type::date(), array(new Date(), new Date(0), new Date(-86400), new Date(86400))),
             array(Type::decimal(), array(new Decimal(1.1), new Decimal(2.2), new Decimal(3.3))),
             array(Type::double(), array(1.1, 2.2, 3.3, 4.4)),
+            array(Type::duration(), array(new Duration(0, 0, 0),
+                                          new Duration(1, 2, 3),
+                                          new Duration(-9223372036854775809, -1, -9223372036854775808))),
```
stamhankar999: I think for line 39, -9223372036854775808 isn't that interesting given months = -9223372036854775809 (the min 64-bit value we support). Maybe make nanos = 0, to make sure we (really C*) accepts non-positive values (as opposed to all positive or all negative only).

stamhankar999:
👍  Aside from my nit about that one test case, lgtm...though CI seems to need some love.

PR #100, opened 2017-02-09
PHP-147 - Bigint creation issue using string equal to PHP_INT_MIN

mikefero:


mpenick:
There's also an issue with large negative hex and octal values. For example:

```php
print new Cassandra\Bigint("-0x8000000000000000") . PHP_EOL; // error
```

Possible fix (you might refine this, this was just to try it out):

```diff
diff --git a/ext/util/math.c b/ext/util/math.c
index 02a05b3..5b38e0f 100644
--- a/ext/util/math.c
+++ b/ext/util/math.c
@@ -169,7 +169,6 @@ php_driver_parse_bigint(char *in, int in_len, cass_int64_t *number TSRMLS_DC)
       base = 2;
       break;
     case 'x':
-      point += 2;
       base = 16;
       break;
     default:
@@ -178,7 +177,7 @@ php_driver_parse_bigint(char *in, int in_len, cass_int64_t *number TSRMLS_DC)
     }
   }

-  if (base == 10) {
+  if (base == 8 || base == 10 || base == 16) {
     point = 0;
   }
   errno = 0;
```

tests/unit/Cassandra/BigintTest.php
```diff
@@ -87,7 +88,9 @@ public function validStrings()
             array("-123", "-123"),
             array("-0123", "-83"),
             array("-0x123", "-291") ,
-            array("-0b1010101", "-85")
+            array("-0b1010101", "-85"),
+            array("" . PHP_INT_MAX, "" . PHP_INT_MAX),
```
avalanche123: you can use explicit casting here:
```php
array((string) PHP_INT_MAX, (string) PHP_INT_MAX),
```

mpenick:
I'm pretty sure this doesn't handle the largest negative binary value too. That said, do not use the possible solution above...

mpenick:
Something like this might be better. Really the only special case is binary so that's what this does.

```diff
diff --git a/ext/util/math.c b/ext/util/math.c
index c845b8f..2fa3a31 100644
--- a/ext/util/math.c
+++ b/ext/util/math.c
@@ -147,11 +147,12 @@ int
 php_driver_parse_bigint(char *in, int in_len, cass_int64_t *number TSRMLS_DC)
 {
   char *end = NULL;
+  cass_uint64_t temp;

   int point = 0;
-  int base = 10;
+  int base = 0; /* Automatically determine the base */

-  /*  Determine the sign of the number. */
+  /* Determine the sign of the number. */
   int negative = 0;
   if (in[point] == '+') {
     point++;
@@ -160,28 +161,30 @@ php_driver_parse_bigint(char *in, int in_len, cass_int64_t *number TSRMLS_DC)
     negative = 1;
   }

-  if (in[point] == '0') {
-    switch(in[point + 1]) {
-    case 'b':
-      point += 2;
-      base = 2;
-      break;
-    case 'x':
-      point += 2;
-      base = 16;
-      break;
-    default:
-      base = 8;
-      break;
-    }
+  /* Handle special case for binary e.g. "0xb0100" */
+  if (in[point] == '0' && in[point + 1] == 'b') {
+    base = 2;
+    point += 2; /* Skip over "0b" */
   }

   errno = 0;
+  temp = (cass_uint64_t) strtoull(in + point, &end, base);

-  *number = (cass_int64_t) strtoll(&(in[point]), &end, base);
-
-  if (negative)
-    *number = *number * -1;
+  if (negative) {
+    if (temp > (cass_uint64_t) INT64_MAX + 1) {
+      errno = ERANGE;
+    } else if (temp == (cass_uint64_t) INT64_MAX + 1) {
+      *number = INT64_MIN;
+    } else {
+      *number = -((cass_int64_t) temp);
+    }
+  } else {
+    if (temp > (cass_uint64_t) INT64_MAX) {
+      errno = ERANGE;
+    } else {
+      *number = temp;
+    }
+  }

   if (errno == ERANGE) {
     zend_throw_exception_ex(php_driver_range_exception_ce, 0 TSRMLS_CC, "Value is too small or too big for bigint: '%s'", in);
```

mpenick:
I think `php_driver_parse_int()` has all the same bugs.

mikefero:
Right now I am working on moving the base determination into another method so that it can be used with `php_driver_parse_int()`, `php_driver_parse_bigint()`, and `php_driver_parse_varint()`. 

I do feel that bumping up to `strtoull` might be overkill since `strtoll` already handles the negative sign in the string. I should have something to commit soon for review.

mpenick:
Using `strtoull` allows parsing a binary representation without copying or overwriting the buffer...

mikefero:
Good call @mpenick. I applied the conversion method; however I made one small modification where if `ERANGE` was set, the max/min value would get associated with the number (similar to `strto*` methods). This allowed the error messages to get more specific for out of bound errors.

I have added the `WIP` label to this ticket in order to address the follwing:

- Add unit tests for `Tinyint` and `Smallint` to validate `php_driver_parse_int ()`
- Move string convert blocks into a macro (thoughts?)

mikefero:
@mpenick @avalanche123 This is good to go for a review again

avalanche123:
:+1: from me

mikefero:
It looks like there are going to be some issues with CI on the Linux side (so far); will re-test and update where applicable. Initial testing was performed on Windows platform using PHP v5.6 (x86).

ext/util/math.c
```diff
@@ -246,7 +223,7 @@ php_driver_parse_varint(char *in, int in_len, mpz_t *number TSRMLS_DC)
 int
 php_driver_parse_decimal(char *in, int in_len, mpz_t *number, long *scale TSRMLS_DC)
 {
-  /*  start is the index into the char array where the significand starts */
+  /*  start is the index into the char array where the significant starts */
```
mpenick: 'significand' was correct.

ext/util/math.h
```diff
@@ -19,6 +19,7 @@
 
 void import_twos_complement(cass_byte_t *data, size_t size, mpz_t *number);
 cass_byte_t *export_twos_complement(mpz_t number, size_t *size);
+int prepare_string_conversion(char *in, int *pos, int *negative);
```
mpenick: This doesn't seem to be used anywhere other than `math.c`. It doesn't need to be expose and could be marked `static`.

tests/unit/Cassandra/TimeTest.php
```diff
@@ -23,6 +23,10 @@
  */
 class TimeTest extends \PHPUnit_Framework_TestCase
 {
+    public function setUp() {
+        date_default_timezone_set("UTC");
```
mpenick: Is this related?

ext/src/Bigint.c
```diff
@@ -25,6 +25,15 @@
 #  define INT64_MIN (-INT64_MAX-1)
 #endif
 
+/* Ensure zend_long is defined for proper conversion */
+#ifndef zend_long
```
mpenick: If `zend_long` isn't defined that means it's PHP version 5.6 or less which means `lval` is a `long`. Source: https://github.com/php/php-src/blob/PHP-5.6/Zend/zend.h#L323 This should be `typedef long zend_long`.

I would consider moving this to `php_driver.h` because this is pretty useful.

ext/src/Bigint.c
```diff
@@ -37,17 +46,7 @@ to_double(zval *result, php_driver_numeric *bigint TSRMLS_DC)
 static int
 to_long(zval *result, php_driver_numeric *bigint TSRMLS_DC)
 {
-  if (bigint->data.bigint.value < (cass_int64_t) LONG_MIN) {
```
mpenick, line 40: I take back what I said, this is a valid check. On a 32-bit system `long`/`zend_long` is a 32-bit signed value so it can't hold the entire range of bigint.

ext/src/Smallint.c
```diff
@@ -242,10 +242,6 @@ PHP_METHOD(Smallint, div)
     }
 
     result->data.smallint.value = self->data.smallint.value / smallint->data.smallint.value;
-    if (result->data.smallint.value * smallint->data.smallint.value != self->data.smallint.value) {
```
mpenick, line 245: I'm curious what this is for... Is there any combination of values that would make this true?

ext/util/math.c
```diff
@@ -89,48 +98,43 @@ int
 php_driver_parse_int(char* in, int in_len, cass_int32_t* number TSRMLS_DC)
 {
   char* end = NULL;
-
-  int point = 0;
-  int base = 10;
-
-  /*  Determine the sign of the number. */
+  int pos = 0;
```
mpenick, line 122: This is a great change.

ext/util/math.c
```diff
@@ -621,3 +598,33 @@ export_twos_complement(mpz_t number, size_t *size)
 
   return bytes;
 }
+
+int
+prepare_string_conversion(char *in, int *pos, int *negative)
```
mpenick: Nice sharing of common logic.

tests/unit/Cassandra/TimeTest.php
mikefero: Not related and can be removed from this PR; was just an issue from time to time while running the unit tests in Windows when the timezone was not configured properly.

ext/src/Smallint.c
mikefero: My thoughts on why this was there was because the documentation indicates the arithmetic operation methods can take a [Numeric](https://github.com/datastax/php-driver/blob/master/ext/doc/Cassandra/Tinyint.php#L71)  value; however it really only takes the same type as the variable instance. Because of that, there is no possible value that would ever make that true except for when `result->data.smallint.value = 0` (e.g. 2/3 = 0 and the conditional would be 0 * 3 != 2) which is actually a bug.

I think we might want to create another ticket that allows the use of `int`|`double`|`string`|`instance_type` because it is a little cumbersome to force the use of just the instance type when the constructors for each type allow for all the other types and handle the conversion.

ext/src/Bigint.c
mikefero: What do you think about creating a new more descriptive type in `php_driver.h` for both PHP 5.6 and PHP 7 that conforms to the way we handle multiple PHP versions?

```
--- a/ext/php_driver.h
+++ b/ext/php_driver.h
@@ -105,6 +105,7 @@ typedef zval php5to7_zval;
 typedef zval *php5to7_zval_args;
 typedef zval *php5to7_zval_arg;
 typedef zend_string *php5to7_string;
+typedef zend_long php5to7_long;
 typedef zend_ulong php5to7_ulong;
 typedef zval php5to7_zend_resource_le;
 typedef zend_resource* php5to7_zend_resource;
@@ -268,6 +269,7 @@ php5to7_string_compare(php5to7_string s1, php5to7_string s2)
 typedef zval *php5to7_zval;
 typedef zval ***php5to7_zval_args;
 typedef char *php5to7_string;
+typedef long php5to7_long;
 typedef ulong php5to7_ulong;
 typedef zend_rsrc_list_entry php5to7_zend_resource_le;
 typedef zend_rsrc_list_entry *php5to7_zend_resource;
```

ext/src/Bigint.c
mikefero: True. Since this is used for the `toInt()` method should we indicate that the `toDouble()` or `toString()` method should be used to get the value contained within the `Bigint`?

mpenick:
+1 squash and merge.

PR #102, opened 2017-02-13
PHP-138 - Deprecate ExecutionOptions in favor of a simple array

stamhankar999:


mpenick:
We could "borrow" the memory of the arguments (instead of copying) when populating the local options e.g. https://github.com/datastax/php-driver/commit/874daaa52c7bfd5e9d43d0ffa8ab978df5e99dde

It eliminates the need to clean up the local options and reduces the code a bit. You can take what you want from this I just wanted to implement something to make sure it worked properly before suggesting it.

stamhankar999:
I think that's a great idea. I was rather annoyed that I have to add this cleanup code all over the place.  Once all the tests pass, I'll make this change.

mpenick:
+1. Look great.

PR #104, opened 2017-02-16
PHP-149

stamhankar999:
- Update tests, docs, and examples to use 32-bit values for Duration attributes


ext/src/Duration.c
```diff
@@ -9,34 +9,31 @@
...
     }
+    *destination = (cass_int32_t) parsed_big_int;
   } else if (Z_TYPE_P(value) == IS_OBJECT &&
              instanceof_function(Z_OBJCE_P(value), php_driver_bigint_ce TSRMLS_CC)) {
     php_driver_numeric *bigint = PHP_DRIVER_GET_NUMERIC(value);
-    *destination = bigint->data.bigint.value;
+    *destination = (cass_int32_t) bigint->data.bigint.value;
   } else {
     throw_invalid_argument(value, param_name, "a long, a double, a numeric string or a " \
                             PHP_DRIVER_NAMESPACE "\\Bigint" TSRMLS_CC);
```
mikefero: Looks like a copy+paste issues, should be duration.

ext/src/Duration.c
```diff
@@ -9,34 +9,31 @@
...
   } else if (Z_TYPE_P(value) == IS_DOUBLE) {
-    *destination = (cass_int64_t) Z_DVAL_P(value);
+    *destination = (cass_int32_t) Z_DVAL_P(value);
   } else if (Z_TYPE_P(value) == IS_STRING) {
-    if (!php_driver_parse_bigint(Z_STRVAL_P(value), Z_STRLEN_P(value), destination TSRMLS_CC)) {
+    cass_int64_t parsed_big_int;
+    if (!php_driver_parse_bigint(Z_STRVAL_P(value), Z_STRLEN_P(value), &parsed_big_int TSRMLS_CC)) {
       return 0;
     }
+    *destination = (cass_int32_t) parsed_big_int;
```
mikefero: If the `Bigint` value is greater than ((2^32) - 1) there will be some wrap around when casting. Should we allow the automatic wrapping?

mikefero (changes requested):

ext/src/Duration.c
stamhankar999: I thought about that a bit, but at the time I figured that Bigint doesn't do any special checks when taking longs/doubles. But now that I think about it, Bigint doesn't suffer with this problem since it's storing its value as a 64-bit number.

So I agree we should catch this overflow and error out, and similarly error out in the clause above where we deal with doubles.

ext/src/Duration.c
stamhankar999: Well, the arguments to the Duration constructor are long, double, or numeric string.

mikefero: Yeah you are right on that, I read it wrong; ignore this comment.

ext/src/Duration.c
```diff
@@ -7,36 +7,69 @@
 
 #include "Duration.h"
 
+#define CASS_INT32_MAX 2147483647L
```
mikefero: I think we should be consistent with the other uses of MAX/MIN in our code base and handle the cases where it is not defined by a particular platform (see [Bigint](https://github.com/datastax/php-driver/blob/e11b36df381d08dd321cdc22e33c7724fe6d6f07/ext/src/Bigint.c#L23-L26))

```
#if !defined(HAVE_STDINT_H) && !defined(_MSC_STDINT_H_)
#  define INT32_MAX 2147483647L
#  define INT32_MIN (-INT32_MAX-1)
#endif
```

tests/unit/Cassandra/DurationTest.php
```diff
@@ -49,6 +49,78 @@ public function testStringArgParseError()
...
+     * @expectedExceptionMessage nanos must be between -2147483648 and 2147483647, -8589934592 given
+     */
+    public function testBigintArgUnderflowError()
+    {
+        new Duration(1, 2, new Bigint("-8589934592"));
+    }
+
+    /**
+     * @expectedException InvalidArgumentException
+     * @expectedExceptionMessageRegExp /days must be between -2147483648 and 2147483647, 8\.?58993.* given/
```
mikefero, line 89: nitpick (consistency): Why the use of regular expressions here and not above?

tests/unit/Cassandra/DurationTest.php
```diff
@@ -49,6 +49,78 @@ public function testStringArgParseError()
...
+     */
+    public function testLongArgOverflowError()
+    {
+        new Duration(1, 8589934592, 2);
+    }
+
+    /**
+     * @expectedException InvalidArgumentException
+     * @expectedExceptionMessageRegExp /days must be between -2147483648 and 2147483647, -8\.?58993.* given/
```
mikefero, line 98: same nitpick here

mikefero (changes requested):

ext/src/Duration.c
stamhankar999: Sounds reasonable; I was looking at constants.hpp in cpp driver as my model.

tests/unit/Cassandra/DurationTest.php
stamhankar999: In this case, the exception message varies among platforms: the "given" is either formatted in scientific notation (on most Windows / VC flavors) or straight out (on Linux).

mikefero: Interesting ... I wonder if it has to do PHP v5.6 using 32-bit longs under the hood and it automatically converts them to a double as the value you are using is larger than PHP_INT_MAX for those platforms.

mikefero:
+1

Recommend rebase since there are some old master commits in this PR

mikefero (approved):

PR #105, opened 2017-02-16
PHP-145 - Add API doc generator system

mpenick:
* Fixed up issues with external API e.g. interfaces that should be
 abstract classes etc.
* Added and corrected yaml files


stamhankar999:
👍  There are very few PHP files in this PR. Does that mean that when generating the php files, they ended up being identical to what we had before? Looks good to merge as long as that's the case.

ext/src/Type.yaml
```diff
@@ -0,0 +1,231 @@
...
+      return:
+        comment: date type
+        type: Cassandra\Type
+    time:
+      comment: Get representation of time type
+      return:
+        comment: time type
+        type: Cassandra\Type
+    duration:
+      comment: Get reperesentation of duration type
```
stamhankar999: reperesentation => representation

ext/doc/generate_doc.php
```diff
@@ -0,0 +1,475 @@
+<?php
+
+define("LICENSE_COMMENT", "/**
+ * Copyright 2015-2016 DataStax, Inc.
```
mikefero: We should probably update this to 2017

ext/src/Custom.yaml
```diff
@@ -0,0 +1,10 @@
+---
+Cassandra\Custom:
+  comment: An interface for representing custom values.
```
mikefero: Updated source to be an abstract class

mikefero (commented):
Couple of nitpicks; looks great overall

ext/doc/Cassandra.php
```diff
@@ -207,17 +263,17 @@
...
      * @see Cassandra\Map::__construct()
      */
-    const TYPE_BLOB = 'blob';
+    const TYPE_BLOB = "blob";
 
     /**
-     * When using a map, collection or set of type boolean, all of its elements
-     * must be booleans.
+     * When using a map, collection or set of type bool, all of its elements
+     * must be bools.
```
stamhankar999: Is this intended to be `bools` or `booleans`?

ext/doc/Cassandra.php
```diff
@@ -160,14 +161,49 @@
...
+    /**
      * When using a map, collection or set of type text, all of its elements
      * must be strings.
      *
      * @see Cassandra\Set::__construct()
      * @see Cassandra\Collection::__construct()
      * @see Cassandra\Map::__construct()
      */
-    const TYPE_TEXT = 'text';
+    const TYPE_TEXT = "text";
```
stamhankar999: Any particular reason to use interpolated (double-quoted) strings instead of single-quoted?

ext/doc/Cassandra/BatchStatement.php
```diff
@@ -37,24 +37,23 @@
...
     /**
      * Adds a statement to this batch.
      *
-     * @param string|Statement $statement string or statement to add
-     * @param array|null $arguments positional or named arguments
      *
      * @throws Exception\InvalidArgumentException
-     *
+     * @param string|Statement $statement string or statement to add
+     * @param array|null $arguments positional or named arguments (optional)
```
stamhankar999: I like the original format better:
```
/**
 * Description
 *
 * @param ...
 * @param ...
 *
 * @throws ...
 *
 * @return ...
 *
 * @see ...
```
Separate each section with a blank line, order the sections roughly as above.

stamhankar999: That said, this can be a later refinement, assuming that doxygen and documentor don't care about the order of sections here.

ext/doc/Cassandra/Bigint.php
```diff
@@ -21,107 +21,97 @@
...
+     * @param Cassandra\Numeric $num a number to add to this one
      * @return Numeric sum
      */
-    public function add(Numeric $addend) {}
+    public function add($num) { }
 
     /**
-     * @param Numeric $subtrahend a number to subtract from this one
-     *
+     * @param Cassandra\Numeric $num a number to subtract from this one
```
stamhankar999: Will qualifying the type names with namespace cause trouble when merging into dse driver repo? I believe doxygen/Documentor handle unqualified types and create proper links.

Conversely, if it is actually incorrect to use unqualified type names, then we should use qualified names.

ext/doc/Cassandra/Cluster/Builder.php
```diff
@@ -23,138 +23,165 @@
...
-    public function withDefaultTimeout($timeout) {}
+    public function withDefaultTimeout($timeout) { }
 
     /**
      * Configures the initial endpoints. Note that the driver will
      * automatically discover and connect to the rest of the cluster.
      *
-     * @param string $host,... one or more ip addresses or hostnames
-     *
+     * @param string $host ,... one or more ip addresses or hostnames
```
stamhankar999: Looks a little weird to have spaces before the comma, but not the end of the world. However, we should have a consistent format for specifying a var-args-style param list. For example, the LineString constructor in the Dse driver has this description:
```
* @param array $points varargs-style array of Point's
```
And the constructor declaration has one `$points` arg, analogous to `withContactPoints` here.

ext/doc/Cassandra/Cluster/Builder.php
```diff
@@ -23,138 +23,165 @@
...
+     * @return Builder self
+     */
+    public function withBlackListHosts($hosts) { }
+
+    /**
+     * Sets the whitelist hosts. Any host not in the whitelist will be ignored
+     * and a connection will not be established. This policy is useful for
+     * ensuring that the driver will only connect to a predefined set of hosts.
+     *
+     * @param string $hosts A comma delimited list of addresses.
```
stamhankar999, line 119: I think this is outright wrong; the [code](https://github.com/datastax/php-driver/blob/cb1023c2525a24634357b4ca4ad44f8c89e8b1cb/ext/src/Cluster/Builder.c#L405-L405) indicates this is actually a var-args-style param.

ext/doc/Cassandra/Cluster/Builder.php
```diff
@@ -23,138 +23,165 @@
...
+     * @return Builder self
+     */
+    public function withDatacenterAwareRoundRobinLoadBalancingPolicy($localDatacenter, $hostPerRemoteDatacenter, $useRemoteDatacenterForLocalConsistencies) { }
+
+    /**
+     * Sets the blacklist hosts. Any host in the blacklist will be ignored and
+     * a conneciton will not be established. This is useful for ensuring that
+     * the driver will not connection to a predefied set of hosts.
+     *
+     * @param string $hosts A comma delimited list of addresses.
```
stamhankar999, line 108: This is actually a var-args style param.

ext/doc/Cassandra/Cluster/Builder.php
```diff
@@ -23,138 +23,165 @@
...
+     */
+    public function withWhiteListHosts($hosts) { }
+
+    /**
+     * Sets the blacklist datacenters. Any datacenter in the blacklist will be
+     * ignored and a connection will not be established to any host in those
+     * datacenters. This policy is useful for ensuring the driver will not
+     * connect to any host in a specific datacenter.
+     *
+     * @param string $dcs A comma delimited list of datacenters.
```
stamhankar999, line 131: varargs, not comma-delimited string

ext/doc/Cassandra/Cluster/Builder.php
```diff
@@ -23,138 +23,165 @@
...
+     * @return Builder self
+     */
+    public function withBlackListDCs($dcs) { }
+
+    /**
+     * Sets the whitelist datacenters. Any host not in a whitelisted datacenter
+     * will be ignored. This policy is useful for ensuring the driver will only
+     * connect to hosts in specific datacenters.
      *
+     * @param string $dcs A comma delimited list of datacenters.
```
stamhankar999: varargs, not comma-delimited string

ext/doc/Cassandra/Cluster/Builder.php
```diff
@@ -23,138 +23,165 @@
...
-    public function withSSL(\Cassandra\SSLOptions $options) {}
+    public function withSSL($options) { }
 
     /**
      * Enable persistent sessions and clusters.
      *
-     * @param bool $enabled whether to enable persistent sessions and clusters
      *                      (optional)
-     *
+     * @param bool $enabled whether to enable persistent sessions and clusters
```
stamhankar999: This looks weird with that stray `(optional)` on the previous line. The generator script did something bad with the multi-line param def.

ext/doc/Cassandra/Cluster/Builder.php
```diff
@@ -166,12 +193,11 @@ public function withPersistentSessions($enabled = true) {}
      *
      * NOTE: Apache Cassandra 3.x supports protocol version 3 and 4 only
      *
-     * @param int $version the actual protocol version, only `1`, `2`, `3`, or
      *                     `4` are supported
-     *
+     * @param int $version the actual protocol version, only `1`, `2`, `3`, or
```
stamhankar999: Same here; the second line got inserted above the first. May be better to update the yaml to just have one line.

ext/doc/Cassandra/Cluster/Builder.php
```diff
@@ -194,70 +219,63 @@ public function withIOThreads($count) {}
...
 
     /**
      * Enables/disables TCP keepalive.
      *
-     * @param float|null $delay the period of inactivity in seconds, after
      *                          which the keepalive probe should be sent over
      *                          the connection. If set to `null`, disables
      *                          keepalive probing.
-     *
+     * @param float|null $delay the period of inactivity in seconds, after
```
stamhankar999: multiline wonkiness again.

ext/doc/Cassandra/Cluster/Builder.php
```diff
@@ -194,70 +219,63 @@ public function withIOThreads($count) {}
...
-    public function withRetryPolicy(RetryPolicy $policy) {}
+    public function withRetryPolicy($policy) { }
 
     /**
      * Sets the timestamp generator.
      *
-     * @param Cassandra\TimestampGenerator $generator A timestamp generator that will be used
      *                                                to generate timestamps for statements.
-     *
+     * @param Cassandra\TimestampGenerator $generator A timestamp generator that will be used
```
stamhankar999: multiline wonkiness again.

ext/doc/Cassandra/Date.php
```diff
@@ -21,36 +21,42 @@
...
      * @return int Absolute seconds from epoch (1970, 1, 1), can be negative
      */
-    public function seconds() {}
+    public function seconds() { }
 
     /**
      * Converts current date to PHP DateTime.
-     *
+     * @param Time $time An optional Time object that is added to the DateTime object.
+     * @return \DateTime PHP representation
```
stamhankar999: Is this referring to a top-level DateTime type? Does it mean to refer to Cassandra\DateTime? Should this be an unqualified ref?

ext/doc/Cassandra/Decimal.php
```diff
@@ -35,96 +35,87 @@
      * $this->assertEquals(21, $decimal->scale());
```
stamhankar999: Why is test assertion code here?

ext/doc/Cassandra/Decimal.php
```diff
@@ -35,96 +35,87 @@
...
-     *
      * @return int Scale
      */
-    public function scale() {}
+    public function scale() { }
 
     /**
-     * @param Numeric $addend a number to add to this one
-     *
+     * @param Cassandra\Numeric $num a number to add to this one
```
stamhankar999: Again, I think we want to use unqualified type names; if so, I'd suggest searching through all yaml files for Cassandra refs and removing them.

ext/doc/Cassandra/DefaultAggregate.php
```diff
@@ -21,68 +21,60 @@
...
+    public function stateType() { }
 
     /**
      * {@inheritDoc}
-     *
-     * @return Cassandra\Type State type of the aggregate
+     * @return Cassandra\Type Return type of the aggregate
      */
-    public function stateType() {}
+    public function returnType() { }
```
stamhankar999: Any particular reason to swap the ordering of stateType and returnType?

ext/doc/Cassandra/DefaultKeyspace.php
```diff
@@ -21,48 +21,97 @@
...
      * {@inheritDoc}
-     *
      * @return array An array of `Cassandra\Table` instances
      */
-    public function tables() {}
+    public function tables() { }
+
+    /**
+     * {@inheritDoc}
+     * @param Type\UserType $name
```
stamhankar999: param type is `string`.

ext/doc/Cassandra/DefaultKeyspace.php
```diff
@@ -21,48 +21,97 @@
...
+     */
+    public function materializedViews() { }
+
+    /**
+     * {@inheritDoc}
+     * @param string $name {@inheritDoc}
+     * @param string|Type $params {@inheritDoc}
+     * @return Function|null {@inheritDoc}
+     */
+    public function function_($name, ...$params) { }
```
stamhankar999: This is proper syntax to express var-args in php 5.6 and later; I think we should standardize on this in our drivers. I'm still not sure what should be written in the `@params` line for such parameters.

ext/doc/Cassandra/DefaultMaterializedView.php
```diff
@@ -21,192 +21,170 @@
...
      */
-    public function option($name) {}
+    public function option($name) { }
 
     /**
      * {@inheritDoc}
-     *
-     * @return array A dictionary of `string` and `Cassandra\Value pairs of the
-     *               view's options.
+     * @return array A dictionary of `string` and `Value` pairs of the
```
stamhankar999: I think if you don't surround these with backticks, documentor might replace Value with a link to its page.

ext/doc/Cassandra/DefaultMaterializedView.php
```diff
@@ -21,192 +21,170 @@
...
+     * @param string $name Name of the column
      * @return Cassandra\Column Column instance
      */
-    public function column($name) {}
+    public function column($name) { }
 
     /**
      * {@inheritDoc}
-     *
      * @return array A list of `Cassandra\Column` instances
```
stamhankar999: Same here; I think if you remove the backtick, Documentor will create links. Don't take my word on this; try it out with one type and see.

ext/doc/Cassandra/Exception/ConfigurationException.php
```diff
@@ -23,4 +23,23 @@
  * invalid because of some configuration issue.
  * For example when attempting to drop a non-existent keyspace.
  */
-class ConfigurationException extends ValidationException {}
+class ConfigurationException extends ValidationException  {
+
+    /**
+     * @param mixed $message
```
stamhankar999, line 29: Can we be more specific about supported types for these params and return types of these methods and those of AuthenticationException above and any other exception classes defined below?

ext/doc/Cassandra/Function.php
```diff
@@ -21,53 +21,54 @@
...
      * Returns the body of the function
      * @return string Body of the function
      */
-    function body();
+    public function body();
 
     /**
-     * Returns whether the function is called when the input columns are null
-     * @return boolean
+     * Determines if a funciton is called when the value is null.
```
stamhankar999: funciton => function

ext/doc/Cassandra/FutureClose.php
```diff
@@ -23,17 +23,17 @@
...
      * {@inheritDoc}
      *
      * @throws Exception\InvalidArgumentException
      * @throws Exception\TimeoutException
      *
-     * @param float|null $timeout
-     *
-     * @return mixed a value that the future has been resolved with
+     * @param int|double|null $timeout A timeout in seconds
+     * @return mixed Nothing
```
stamhankar999: If it returns nothing, why have an @return?

ext/doc/Cassandra/Index.php
```diff
@@ -21,48 +21,50 @@
...
      * @return string Class name of a custom index
      */
-    function className();
+    public function className();
 
     /**
-     * Returns `true` if this is a custom index
-     * @return boolean
+     * Determines if the index is a custom index.
+     * @return bool Returns `true` if this is a custom index
```
stamhankar999: Since this is an @return, it is redundant to say `Returns...`.

ext/doc/Cassandra/MaterializedView.php
```diff
@@ -21,11 +21,170 @@
...
+    /**
+     * Return a table's option by name
+     * @param string $name The name of the option
+     * @return Cassandra\Value Value of an option by name
+     */
+    public abstract function option($name);
+
+    /**
+     * Returns all the table's options
+     *               view's options.
```
stamhankar999: Description wonkiness.

ext/doc/Cassandra/MaterializedView.php
```diff
@@ -21,11 +21,170 @@
...
+
     /**
      * Returns the base table of the view
      * @return Cassandra\Table Base table of the view
      */
-    function baseTable();
+    public abstract function baseTable();
+
+    /**
+     * Returns the name of this table
```
stamhankar999: view, not table. You should read through all of the yaml for this type and correct all refs.

ext/doc/Cassandra/MaterializedView.php
```diff
@@ -21,11 +21,170 @@
...
+     * Return a table's option by name
+     * @param string $name The name of the option
+     * @return Cassandra\Value Value of an option by name
+     */
+    public abstract function option($name);
+
+    /**
+     * Returns all the table's options
+     *               view's options.
+     * @return array A dictionary of `string` and `Value` pairs of the
```
stamhankar999: Maybe remove backticks, if documentor will replace with links.

ext/doc/Cassandra/RetryPolicy/Logging.php
```diff
@@ -18,18 +18,17 @@
...
-final class Logging implements RetryPolicy
-{
+final class Logging implements RetryPolicy {
+
     /**
      * Creates a new Logging retry policy.
      *
-     * @param Cassandra\RetryPolicy $childPolicy Any retry policy other than
      *                                           Cassandra\Logging
+     * @param Cassandra\RetryPolicy $childPolicy Any retry policy other than
```
stamhankar999: Wonky multiline param description

ext/doc/Cassandra/Rows.php
```diff
@@ -21,135 +21,135 @@
 /**
  * Rows represent a result of statement execution.
  */
-final class Rows implements \Iterator, \Countable, \ArrayAccess
-{
+final class Rows implements \Iterator, \ArrayAccess {
```
stamhankar999, line 24: Why not implement the \Countable interface?

ext/doc/Cassandra/Session.php
```diff
@@ -20,93 +20,74 @@
 
 /**
  * A session is used to prepare and execute statements.
- *
- * @see Cluster::connect()
```
stamhankar999: Why remove the @see on this?

ext/doc/Cassandra/Session.php
```diff
@@ -20,93 +20,74 @@
...
-interface Session
-{
+interface Session {
+
     /**
-     * Returns current performance/diagnostic metrics.
+     * Execute a query.
      *
-     * @return array Performance/Diagnostic metrics.
+     * @throws Exception
```
stamhankar999: Again, I don't like the order of sections here, but I'm surprised that there are blank lines above and below @throws. From earlier php code, I thought the code generator doesn't add those. That said, I would like a blank line between each section.

ext/doc/Cassandra/Session.php
```diff
@@ -20,93 +20,74 @@
...
+     *
+     * @param string|Statement $statement string or statement to be executed.
+     * @param array|ExecutionOptions|null $options execution options (optional)
+     * @return Rows A colleciton of rows.
```
stamhankar999: colleciton -> collection

ext/doc/Cassandra/Session.php
```diff
@@ -20,93 +20,74 @@
...
 
     /**
-     * Returns current schema.
-     *
-     * NOTE: the returned Schema instance will not be updated as the actual
-     *       schema changes, instead an updated instance should be requested by
-     *       calling Session::schema() again.
-     *
-     * @return Schema current schema.
+     * Execute a query asyncrhonously. This method returns immediately, but
```
stamhankar999: asyncrhonously -> asynchronously

ext/doc/Cassandra/Table.php
```diff
@@ -21,163 +21,164 @@
...
      * @return Cassandra\Value Value of an option by name
      */
-    function option($name);
+    public function option($name);
 
     /**
      * Returns all the table's options
-     * @return array A dictionary of `string` and `Cassandra\Value pairs of the
      *               view's options.
+     * @return array A dictionary of `string` and `Value` pairs of the
```
stamhankar999: Possibly remove backticks.

ext/doc/Cassandra/Timeuuid.php
```diff
@@ -21,55 +21,50 @@
...
      */
-    public function version() {}
+    public function version() { }
 
     /**
      * Unix timestamp.
      *
-     * @return int seconds
      * @see time
+     * @return int seconds
```
stamhankar999: I think @return should come before @see. See the template I wrote up in one of the very early comments.

ext/doc/Cassandra/Type.php
```diff
@@ -146,47 +204,42 @@
...
     /**
-     * Instantiate a value of this type from provided value(s).
+     * Initialize a user type
+     * @code{.php}
+     * <?php
+     * use Cassandra\Type;
      *
-     * @throws Exception\InvalidArgumentException when values given cannot be
-     *                                            represented by this type.
+     * $userType = Type::userTupe("a", Type::varchar(), "b", Type::int())
```
stamhankar999: userTupe => userType

ext/doc/Cassandra/Type/UserType.php
```diff
@@ -18,46 +18,61 @@
...
+/**
+ * A class that represents a user type. The user type is able to represent a
+ * composite type of one or more types accessed by name.
+ */
+final class UserType extends Type {
+
+    private function __construct() { }
+
+    /**
+     * @param mixed $name
```
stamhankar999: should be string, not mixed.

ext/doc/Cassandra/Type/UserType.php
```diff
@@ -18,46 +18,61 @@
...
-{
     /**
      * Returns type name for the user type
      * @return string
      */
-    public function name() {}
+    public function name() { }
+
+    /**
+     * @param mixed $keyspace
```
stamhankar999: string, not mixed

ext/doc/Cassandra/Type/UserType.php
```diff
@@ -18,46 +18,61 @@
...
+     * no values given, creates an empty user type.
      *
      * @throws Exception\InvalidArgumentException when values given are of a
      *                                            different types than what the
      *                                            user type expects.
-     *
-     * @param  mixed $value,...      One or more name/value pairs to be added to the user type.
-     *                               When no values given, creates an empty user type.
-     * @return Cassandra\UserTypeValue  A user type value with given name/value pairs.
+     * @param mixed $value ,...      One or more name/value pairs to be added to the user type.
```
stamhankar999: Zero or more, right? [code](https://github.com/datastax/php-driver/blob/cb1023c2525a24634357b4ca4ad44f8c89e8b1cb/ext/src/Type/UserType.c#L190-L190)

ext/doc/generate_doc.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+php -d  extension=cassandra.so -d extension_dir=../modules/ generate_doc.php ..
```
stamhankar999: This is written assuming cwd is `ext/doc`, while the script in the dse driver assumes cwd is `ext`. I think this script should work no matter what cwd is. Something like:
```
#!/bin/sh
DIR=`dirname "$0"`
php -d extension=cassandra.so -d extension_dir="$DIR/../modules" "$DIR/generate_doc.php" $DIR/..
```

ext/doc/generate_doc_yaml.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+php -d extension=cassandra.so -d extension_dir=../modules/ generate_doc_yaml.php ..
```
stamhankar999: Same idea as generate_doc.sh

ext/doc/Cassandra/BatchStatement.php
mpenick: Agreed. The original format is better. This was an unintended  side effect.

ext/doc/Cassandra/Bigint.php
mpenick: I'd prefer to use the unqualified. Good question.

ext/doc/generate_doc.sh
mpenick: 👍

ext/doc/Cassandra.php
mpenick: Changing it to single quotes.

ext/doc/Cassandra/Cluster/Builder.php
mpenick: Agreed, we've never handled our flexible/vararg arguments well in the API docs. I'll need to think on this.

ext/doc/Cassandra/Date.php
mpenick: > Converts current date to PHP DateTime.

ext/doc/Cassandra/DefaultAggregate.php
mpenick: It's based on the order in the yaml file. Not intentional...

ext/doc/Cassandra/Exception/ConfigurationException.php
mpenick: This is outside the scope of what I'm trying to do here.

mpenick:
Merged at 922e37a39b2505e4f62fa1aeb75eab0d7df2b4b0

ext/doc/Cassandra/Float.php
```diff
@@ -21,107 +21,118 @@
...
  */
-final class Float implements Value, Numeric
-{
-    /**
-     * Minimum possible Float value
-     *
-     * @return Float minimum value
-     */
-    public static function min() {}
+final class Float_ implements Value, Numeric {
```
qkdreyer, line 24: where does this class name change comes from ?

qkdreyer: [PSR-4](https://www.php-fig.org/psr/psr-4/) recommends having file name match class name. Shouldn't this file be renamed to `ext/doc/Cassandra/Float_.php` ?

PR #107, opened 2017-02-21
PHP-151 - Smallint, Tinyint, and Duration constructors should error out with a RangeException for out-of-range parameters

stamhankar999:
All numeric type constructors should error out with RangeException for out-of-range parameters.

### DIFF
--- a/ext/util/math.c
+++ b/ext/util/math.c
@@ -198,7 +198,7 @@ php_driver_parse_bigint(char *in, int in_len, cass_int64_t *number TSRMLS_DC)
 
   if (errno == ERANGE) {
     zend_throw_exception_ex(php_driver_range_exception_ce, 0 TSRMLS_CC,
-      "value must be between %lld and %lld, %s given", INT64_MIN, INT64_MAX, in);
+      "value must be between " LL_FORMAT " and " LL_FORMAT ", %s given", INT64_MIN, INT64_MAX, in);
     return 0;
   }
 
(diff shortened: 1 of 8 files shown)


ext/util/math.c
```diff
@@ -200,11 +197,8 @@ php_driver_parse_bigint(char *in, int in_len, cass_int64_t *number TSRMLS_DC)
   }
 
   if (errno == ERANGE) {
-    if (*number == INT64_MAX) {
-      zend_throw_exception_ex(php_driver_range_exception_ce, 0 TSRMLS_CC, "Value is too big for bigint: '%s'", in);
-    } else {
-      zend_throw_exception_ex(php_driver_range_exception_ce, 0 TSRMLS_CC, "Value is too small for bigint: '%s'", in);
-    }
+    zend_throw_exception_ex(php_driver_range_exception_ce, 0 TSRMLS_CC,
+      "value must be between %lld and %lld, %s given", INT64_MIN, INT64_MAX, in);
```
mpenick: Why "Value" --> "value"?

stamhankar999: The arg name for all of the number class constructors is 'value'. So, making this lower-case and making it consistent with range error messages in the various types seemed like a good thing.

mpenick: That makes sense.

mpenick:
+1 Great work. Ship it!

mpenick:
I like the `LL_FORMAT` stuff.

PR #108, opened 2017-02-23
PHP-152 - Add duration support to composite types

mpenick:


mikefero:
+1

Tested and will need to make updates to the integration tests for Map (duration is not allowed as a key) and Set (duration is not allowed in sets)

mpenick:
Merged at b61f333cecd97af5faa59eda7faa743fd5050b97

PR #110, opened 2017-03-28
Fix platform deps

dzuelke:
The package doesn't declare its `ext-cassandra` dependencies in `require`, only in `require-dev`.


dzuelke:
Nah.

dzuelke:
Trivial fix, you can just replicate it yourselves :)

mpenick:
Thanks. Sorry about the CLA.

PR #111, opened 2017-04-28
v1.3.1 - Merge duration and documentation fixes

mpenick:


features/README.md
```diff
@@ -533,4 +531,4 @@ The PHP Driver follows the architecture of [the C/C++ Driver](http://datastax.gi
 
 ### Persistent sessions
```
mikefero, line 532: I feel we can remove this section entirely from the `Architecture` portion as `Persistent Sessions` are discussed at the beginning of this documentation.

mikefero (approved):
Merge looks good; looks like the cpp-driver submodule needs to be updated to point to the latest version in master for acceptable cpp-driver dependency version (that can wait until after cpp-driver release to point to the release tag if desired).

PR #112, opened 2017-05-18
PHP-189 - Map compare doesn't compare values

mpenick:
`php_driver_map_compare()` doesn't compare values so if two maps have
the same keys, but different values they will compare as being equal.

### DIFF
--- a/ext/src/Map.c
+++ b/ext/src/Map.c
@@ -518,11 +518,14 @@ php_driver_map_compare(zval *obj1, zval *obj2 TSRMLS_DC)
   }
 
   HASH_ITER(hh, map1->entries, curr, temp) {
-    php_driver_map_entry *entry;
+    php_driver_map_entry *entry = NULL;
     HASH_FIND_ZVAL(map2->entries, PHP5TO7_ZVAL_MAYBE_P(curr->key), entry);
     if (entry == NULL) {
       return 1;
     }
+    result = php_driver_value_compare(PHP5TO7_ZVAL_MAYBE_P(curr->value),
+                                      PHP5TO7_ZVAL_MAYBE_P(entry->value));
+    if (result != 0) return result;
   }
 
   return 0;
--- a/ext/src/Map.c
+++ b/ext/src/Map.c
@@ -513,8 +513,8 @@ php_driver_map_compare(zval *obj1, zval *obj2 TSRMLS_DC)
   result = php_driver_type_compare(type1, type2 TSRMLS_CC);
   if (result != 0) return result;
 
-  if (HASH_COUNT(map1->entries) != HASH_COUNT(map1->entries)) {
-   return HASH_COUNT(map1->entries) < HASH_COUNT(map1->entries) ? -1 : 1;
+  if (HASH_COUNT(map1->entries) != HASH_COUNT(map2->entries)) {
+   return HASH_COUNT(map1->entries) < HASH_COUNT(map2->entries) ? -1 : 1;
   }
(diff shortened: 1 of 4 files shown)


ext/src/Map.c
```diff
@@ -518,11 +518,14 @@ php_driver_map_compare(zval *obj1, zval *obj2 TSRMLS_DC)
   }
```
stamhankar999, line 518: This clause has a typo; it should be comparing map1 to map2!

ext/src/Map.c
```diff
@@ -518,11 +518,14 @@ php_driver_map_compare(zval *obj1, zval *obj2 TSRMLS_DC)
...
 
   HASH_ITER(hh, map1->entries, curr, temp) {
-    php_driver_map_entry *entry;
+    php_driver_map_entry *entry = NULL;
     HASH_FIND_ZVAL(map2->entries, PHP5TO7_ZVAL_MAYBE_P(curr->key), entry);
     if (entry == NULL) {
       return 1;
     }
```
stamhankar999, line 525: It makes me a little nervous that there's no circumstance where we'll return -1 for a key in map2 that is missing in map1. Also, does HASH_ITER iterate the map entries in a deterministic way? I think this compare function does a great job of distinguishing equality/inequality. But I don't think it would be safe to be used for sorting a collection of maps (which may be more a theoretical concern than practical, but you never know).

Example:
map1 has one entry: k1, v1
map2 has one entry: k2, v2

php_driver_map_compare(map1, map2) => 1
php_driver_map_compare(map2, map1) => 1

I think in php-land that translates to `map1 < map2` and `map2 < map1` both being true, which make maps ill-suited to be stored in sorted collections (e.g. a binary search of such a collection may not find a map).

Upon further reflection, a key existing in map1 that does *not* exist in map2 implies that there is a key in map2 that doesn't exist in map1, since map sizes are verified to be the same by the time we reach this part of the code. So then the question boils down to which "missing key" is greater/less than the other "missing key" to produce a stable ordering/comparison of the maps. `php_driver_value_compare`s of missing keys would resolve that.

It sounds expensive, but it would be more correct.

stamhankar999 (changes requested):

ext/src/Map.c
mpenick: Yeah, I thought about this when originally implementing it and favored the fast and easy to implement route. I think that documenting the limitation might be a better route here. What do you think?

mpenick: > Also, does HASH_ITER iterate the map entries in a deterministic way?

No, it uses insertion order.

stamhankar999: Yeah, let's just document it.

mpenick: For better or worse. I just discovered that PHP's internal hash table does the same thing for arrays with keys: https://github.com/php/php-src/blob/master/Zend/zend_hash.c#L2352-L2364

```php
array("a" => 1) <=> array("b" => 1);
# => 1
array("b" => 1) <=> array("a" => 1);
# => 1
```

ext/src/Map.c
mpenick: Good catch!

ext/src/Map.c
stamhankar999: Very interesting.

stamhankar999 (approved):

mpenick:
Merged at 63935137c3d7a873d9443335345bd718b2454379

PR #113, opened 2017-06-16
Timeuuid can be initialized from UUID string

zacek:
The current \Cassandra\Timeuuid object has one drawback - it cannot be reconstructed from the string. The C driver supports this but not the PHP extension. In our project we are passing Timeuuid as string via gearman job and this missing functionality is a big problem for us.

Please consider extending the object to support it.

### DIFF
--- a/tests/unit/Cassandra/TimeUuidTest.php
+++ b/tests/unit/Cassandra/TimeUuidTest.php
@@ -54,4 +54,32 @@ class TimeUuidTest extends \PHPUnit_Framework_TestCase
             array(new TimeUuid(0), new TimeUuid(2))
         );
     }
+
+    /**
+     * TimeUuid can be created from string
+     */
+    public function testInitFromStringType1()
+    {
+        new TimeUuid('5f344f20-52a3-11e7-915b-5f4f349b532d');
+    }
+
+    /**
+     * TimeUuid cannot be created from UUID type 4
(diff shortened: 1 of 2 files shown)


zacek:
Could you give me some feedback, please? Usually you are responding very quickly. Is there something wrong with the pull request or is it just the summer time?
Thanks.

tests/unit/Cassandra/TimeUuidTest.php
```diff
@@ -54,4 +54,32 @@ public function notEqualTypes()
...
+
+    /**
+     * TimeUuid cannot be created from invalid string
+     * @expectedException         Cassandra\Exception\InvalidArgumentException
+     * @expectedExceptionMessage  Invalid UUID
+     */
+    public function testInitFromInvalidString()
+    {
+        new TimeUuid('invalid');
+    }
```
stamhankar999, line 84: Can you add a test case for constructing with a different type of object (say a `DateTime`) to verify the `default` case of your switch?

stamhankar999 (changes requested):
I think this looks very good. Just add the requested test and we'll go from there. I'm not an expert on uuid's so I'd like @mpenick to also check this out and make sure he's happy with it.

zacek:
done, test added

stamhankar999 (approved):

PR #114, opened 2017-08-02
PHP-191: Updates to README and copyright dates

mikefero:

PR #116, opened 2017-08-03
PHP-117 - Fix broken links

mpenick:


docs.yaml
```diff
@@ -35,9 +35,15 @@ links:
...
+rewrites:
+  - http://downloads.datastax.com/php-driver/1.0.0/: http://downloads.datastax.com/php-driver/
+  - http://downloads.datastax.com/php-driver/1.1.0/: http://downloads.datastax.com/php-driver/
+  - http://downloads.datastax.com/php-driver/1.2.0/: http://downloads.datastax.com/php-driver/
+  - http://datastax.github.io/php-driver/api/Cassandra/interface.Type/: https://docs.datastax.com/en/developer/php-driver/latest/api/Cassandra/class.Type/
+  - http://datastax.github.io/php-driver/features/http://localhost:8000/features/datatypes/: https://docs.datastax.com/en/developer/php-driver/latest/features/
 versions:
   - name: 1.3
-    ref: v1.3.1
+    ref: HEAD
```
mikefero, line 46: Are any of these changes required? It seems to me only the `features/README.md` is the only thing that needed to be fixed,

features/README.md
```diff
@@ -483,7 +483,7 @@ foreach ($schema->keyspaces() as $keyspace) {
 
 ### Data Types
 
-The PHP driver for Apache Cassandra supports [a variety of datatypes](/features/core/datatypes/).
+The PHP driver for Apache Cassandra supports [a variety of datatypes](/features/datatypes/).
```
mikefero, line 486: This should be `/features/datatypes.feature` in order to work in GitHub. The rewrite rule can be ignored as that was addressed in #114.

mikefero (changes requested):
Couple of comments and a question.

docs.yaml
mpenick: Yes, this fixes links in documentation for older versions of the API.

features/README.md
mpenick: The issue is the documentation has many, many links that don't work from GitHub markup, but work properly on doc.datastax.com. Doing this would make it the only exception. :) 

Here's a follow up story to address the bigger issue: https://datastax-oss.atlassian.net/browse/PHP-199

docs.yaml
mikefero: Nice!

mikefero (approved):

PR #118, opened 2017-08-04
PHP 198 - Merge DSE fixes to core 

mpenick:
* PHP-118 - Unable to bind list values using global type constants
* PHP-172 - php_driver_value_hash hashes IS_TRUE and IS_FALSE to 1 in PHP7
* PHP-163 - Disable hostname resolution if not supported
* PHP-142 - Memory Leak in Timestamp::toDateTime()


mikefero (approved):
Ship it!

PR #121, opened 2017-09-29
Static build

Skunnyk:
This fix the static build of the extension inside php


ext/config.m4
```diff
@@ -2,8 +2,20 @@ PHP_ARG_WITH(cassandra, Enable Cassandra extension,
...
 
 AC_MSG_CHECKING([for supported PHP version])
-PHP_CASSANDRA_FOUND_PHP_VERSION=`${PHP_CONFIG} --version`
+if test -z "$PHP_VERSION"; then
+  if test -z "$PHP_CONFIG"; then
+    AC_MSG_ERROR([php-config not found])
+  fi
+  PHP_CASSANDRA_FOUND_PHP_VERSION=`$PHP_CONFIG --version`
+else
+  PHP_CASSANDRA_FOUND_PHP_VERSION=`echo "$PHP_VERSION"`
```
mikefero: No reason to `echo` here; simply assign the `PHP_CASSANDRA_FOUND_PHP_VERSION` variable.

ext/php_driver.h
```diff
@@ -460,6 +460,9 @@ php5to7_string_compare(php5to7_string s1, php5to7_string s2)
 
 extern zend_module_entry php_driver_module_entry;
 
+#define cassandra_module_ptr &php_driver_module_entry
+#define phpext_cassandra_ptr cassandra_module_ptr
```
mikefero: Could you please update `#define cassandra_module_ptr &php_driver_module_entry` => **`#define phpext_cassandra_ptr &php_driver_module_entry`** and then remove **`#define phpext_cassandra_ptr cassandra_module_ptr`**.

mikefero (changes requested):
Thanks for the contribution and nice catch! Just a couple of requested changes.

Skunnyk:
Thank you for the review. 
PR updated

mikefero:
Thank you for the quick turnaround. The failing tests on TravisCI are nothing to worry about right now; a fix for that will come in later in the week.

Squashing and merging PR.

PR #122, opened 2017-10-05
PHP-204: update build.yaml to include a connection test

ArunTejCh:


build.yaml
```diff
@@ -102,7 +102,24 @@ release:
...
                   PHP_TOKENS=($(echo ${PHP_VERSION} | tr "-" " "))
-                  sudo ${PACKAGE_UTILITY_REMOVE} ${PHP_TOKENS[0]}-common
+                  # Installing packaging and testing connection
+                  sudo ${PACKAGE_INSTALL} packages/${PHP_TOKENS[0]}*.${PACKAGE_TYPE}
+                  php <<EOF
+                  <?php
+                  \$cluster = Cassandra::cluster()->build();
+                  \$keyspace = 'system';
+                  try{
+                    \$session = \$cluster->connect(\$keyspace);
```
mpenick: The keyspace is optional. It's not needed.

(Ref: https://docs.datastax.com/en/developer/php-driver/1.3/api/Cassandra/interface.Cluster/#method-connect)

mpenick: The `$session` variable is never used. It's not needed.

build.yaml
```diff
@@ -102,7 +102,24 @@ release:
                   find build -type f -name "*.${PACKAGE_TYPE}" -exec mv {} packages \; 
                   rm -rf build
                   PHP_TOKENS=($(echo ${PHP_VERSION} | tr "-" " "))
-                  sudo ${PACKAGE_UTILITY_REMOVE} ${PHP_TOKENS[0]}-common
+                  # Installing packaging and testing connection
+                  sudo ${PACKAGE_INSTALL} packages/${PHP_TOKENS[0]}*.${PACKAGE_TYPE}
+                  php <<EOF
+                  <?php
+                  \$cluster = Cassandra::cluster()->build();
```
mpenick: For the sake of simplicity and  to reduce escaping you could remove the `$cluster` temporary variable.

build.yaml
```diff
@@ -102,7 +102,24 @@ release:
...
+                  \$keyspace = 'system';
+                  try{
+                    \$session = \$cluster->connect(\$keyspace);
+                  } catch (Cassandra\\Exception\\RuntimeException \$e) {
+                    echo 'Caught exception: ',  \$e->getMessage(), "\n";
```
mpenick: You could do something like `echo "Success" . PHP_EOL"` here and then check for that output in the bash if statement below.  If you're extra paranoid you could check `$e->getMessage()` for the particular error you were expecting and only echo success in that case, but I think that's overkill as we've proven that we've successfully loaded the module by virtue of throwing/catching the `Cassandra\Exception\RuntimeException`.

build.yaml
```diff
@@ -102,7 +102,24 @@ release:
...
+                  }     
+            EOF
+                  # Throw error if connection test fails
+                  if [ $? -ne 0 ] ; then
```
mpenick: I'd would like this to check the output of the test script above. It's sort of paranoid, but I want to prove that the actual code ran and exit codes can't fully do that because it's more likely that `php` falsely returns a 0 exit code than randomly prints "Success".

mpenick (changes requested):

build.yaml
```diff
@@ -102,7 +102,22 @@ release:
                   find build -type f -name "*.${PACKAGE_TYPE}" -exec mv {} packages \; 
                   rm -rf build
                   PHP_TOKENS=($(echo ${PHP_VERSION} | tr "-" " "))
-                  sudo ${PACKAGE_UTILITY_REMOVE} ${PHP_TOKENS[0]}-common
+                  # Installing packaging and testing connection
+                  sudo ${PACKAGE_INSTALL} packages/${PHP_TOKENS[0]}*.${PACKAGE_TYPE}
```
mikefero: Overkill, but i would like to see a check for the module being loaded before the script is executed; this would also give you a short circuit mechanism as you would know that the module didn't load so therefore the PHP script won't work.

build.yaml
```diff
@@ -102,7 +102,22 @@ release:
...
+                  php <<EOF
+                  <?php
+                  try{
+                    Cassandra::cluster()->build()->connect();
+                  } catch (Cassandra\\Exception\\RuntimeException \$e) {
+                    echo 'Caught exception: ',  \$e->getMessage(), "\n";
+                  }     
+            EOF
```
mikefero: nit: indent

mikefero (changes requested):
Couple of comments, overall it looks good (once @mpenick comments are taken care of). Another thing I would like to see would be the removal of the looping through the PHP versions and moving that into the matrix instead. That way we can build in parallel and not fail fast on just one version of PHP should it be the culprit.

build.yaml
ArunTejCh: EOF doesnt work if the indent is deeper. Weird!

build.yaml
ArunTejCh: Good idea. I'm facing an issue with module not loading on php-dse-driver. This would help identify such issues.

build.yaml
```diff
@@ -102,7 +104,31 @@ release:
...
+                  # Installing packaging and testing connection
+                  sudo ${PACKAGE_INSTALL} packages/${PHP_TOKENS[0]}*.${PACKAGE_TYPE}
+
+                  # Checking if cassandra php module is installed
+                  $PHP_EXEC -m | grep -q cassandra
+                  if [ $? -ne 0 ] ; then
+                    echo "cassandra php module not installed. Marking build as failure"
+                    exit 1
+                  fi 
+                  cat > $TEST_FILE <<EOF
```
mpenick: This could be simplified:
```php
PROGRAM_OUTPUT=$(php <<EOF
<?php
# script here
EOF)
if [ $PROGRAM_OUTPUT != "Success" ]
  # Error
fi
```

build.yaml
```diff
@@ -27,6 +27,8 @@ release:
             RELEASE=${TOKENS[1]}
             CPP_DRIVER_VERSION=2.7.0
             CPP_DRIVER_SUBVERSION=1
+            TEST_FILE=test.php
```
mpenick: This is not needed if you make the simplification below.

mpenick:
👍 squash the commits and I'll merge.

PR #125, opened 2018-06-25
Updating cpp driver dependencie

sandvige:
We need the php-driver on Debian 9, which is shipped with OpenSSL 1.1 by default. The cpp-driver is not compatible with OpenSSL 1.1 before 2.8.0

https://datastax-oss.atlassian.net/browse/CPP-438


mikefero:
Thanks @sandvige; landed 8f05497.

PR #126, opened 2018-06-28
Issue format

remicollet:
Notice, those warning are strangely not raised against PHP < 7.3, but the problem stil exists.

There still tons of -Wformat issues related to %lld vs %ld.

This PR only fix critical ones.


chikymanzanares:
it is possible install the driver for PHP 7.3?

PR #133, opened 2019-06-04
Potential overflow with UDT string type.

joeyhub:
PHP strings are not guaranteed null terminated. I'm not sure if you terminate it pre-emptively. If not, depending on engine internals, which might not be static, you can overrun looking for that null byte. Additionally if it already has the information it needs, no reason not to use it.

This might cause other issues down the pipeline if there's no check for null bytes up to the length, in that case it would need to be null terminated ahead of use instead.

### DIFF
--- a/ext/util/collections.c
+++ b/ext/util/collections.c
@@ -593,7 +593,7 @@ php_driver_user_type_set(CassUserType *ut,
   case CASS_VALUE_TYPE_TEXT:
   case CASS_VALUE_TYPE_ASCII:
   case CASS_VALUE_TYPE_VARCHAR:
-    CHECK_ERROR(cass_user_type_set_string_by_name_n(ut, name, Z_STRVAL_P(value), Z_STRLEN_P(value)));
+    CHECK_ERROR(cass_user_type_set_string_by_name_n(ut, name, strlen(name), Z_STRVAL_P(value), Z_STRLEN_P(value)));
     break;
   case CASS_VALUE_TYPE_BIGINT:
   case CASS_VALUE_TYPE_COUNTER:
--- a/ext/util/collections.c
+++ b/ext/util/collections.c
@@ -593,7 +593,7 @@ php_driver_user_type_set(CassUserType *ut,
   case CASS_VALUE_TYPE_TEXT:
   case CASS_VALUE_TYPE_ASCII:
   case CASS_VALUE_TYPE_VARCHAR:
-    CHECK_ERROR(cass_user_type_set_string_by_name(ut, name, Z_STRVAL_P(value)));
+    CHECK_ERROR(cass_user_type_set_string_by_name_n(ut, name, Z_STRVAL_P(value), Z_STRLEN_P(value)));
     break;
   case CASS_VALUE_TYPE_BIGINT:
   case CASS_VALUE_TYPE_COUNTER:


joeyhub:
Made this because I can't create an issue. IF this is a problem is seems likely to be the case only with the UDT type. When the type is exactly varchar/string, it uses _n so I think this is one where it can slip through the cracks. I guess someone should test UDT types with string and see if they can get a segfault.

ext/util/collections.c
```diff
@@ -593,7 +593,7 @@ php_driver_user_type_set(CassUserType *ut,
   case CASS_VALUE_TYPE_TEXT:
   case CASS_VALUE_TYPE_ASCII:
   case CASS_VALUE_TYPE_VARCHAR:
-    CHECK_ERROR(cass_user_type_set_string_by_name(ut, name, Z_STRVAL_P(value)));
+    CHECK_ERROR(cass_user_type_set_string_by_name_n(ut, name, strlen(name), Z_STRVAL_P(value), Z_STRLEN_P(value)));
```
mikefero, line 596: `name` is being retrieved from the zend hash map and is [guaranteed to be a null terminated string](http://www.phpinternalsbook.com/php7/internal_types/strings/zend_strings.html#php-s-history-and-classical-c-strings); therefore the `cass_user_type_set_string_by_name_n()` is unnecessary.

mikefero (commented):
@joeyhub The issues in this repository have been disabled which is why you are not able to create a new issue. Below are the following places you can go to create and/or discuss issues:

* JIRA: https://datastax-oss.atlassian.net/browse/PHP
* Mailing List: https://groups.google.com/a/lists.datastax.com/forum/#!forum/php-driver-user
* DataStax Academy via Slack: https://academy.datastax.com/slack

ext/util/collections.c
joeyhub: Value is user defined as far as I can tell. Generally speaking if that's the case, php strings are usually null terminated though it's not strictly guaranteed to be the case.

If it's coming from outside, the question is what the PHP intention is. Are people writing php C meant to always set it to 0, for example when referencing a string somewhere else (using strdup if necessary) or is there the potential to have it set to something else. Generally speaking guides on PHP say it's just there for a memory hack.

The original intention of always having +1 byte was apparently a hack. It's a mystery if it's always meant to be 0 and to provide safety (at least against overrunning).

If it is an issue, it would take a rare combination of those two circumstances to cause a problem.

joeyhub:
I think the test is failing for this for some other reason as another push request has the same failure? Might just be coincidence it's UDF.

Looking around, generally speaking PHP strings internally virtually always initialize the hack/padding byte at the end of the string sequence (though I don't think that's so in PHP < 7 which just has a char * rather than combined struct). PHP7s combined struct I think prevents people reusing memory as much as well but won't entirely rule out other extensions doing things with strings or oversights in the core that might result in an odd byte at the end. There's no actual guarantee anywhere though that it should be nul though. There's no instruction to extension writers to set it like that. It's not specified anywhere, it's just implemented like that in the core that it's initialized to 0. My personal opinion of security impact is that it would be very low in general, only rare cases might have problems.

However, if code is to be hardened then better to use _n since the length is already available. Also makes for consistency and saves on a pointless strlen check,

PR #142, opened 2020-02-20
WIP: Strip objects optimisation

joeyhub:
This is a work in progress for removing object wrappage. This is the minimal I need for my use case but not yet appropriate for other use as it breaks BC and needs lib redesign to work with PHP properly.

The problem is other OOP languages with strict typing, etc such as C++, C# and Java, can probably box and unbox things but not so with PHP so the wrappage is a guaranteed RAM/CPU issue with the only want to optimise it to outright remove it.


joeyhub:
To fix this more holistically then I probably want to wipe out the objects altogether / put on another layer and just use type + value with object wrapper being some other option or something.

In some cases the user can happily avoid type because defaults / implicits and knowns will do the trick.

ext/config.m4
```diff
@@ -268,19 +267,19 @@ if test "$PHP_CASSANDRA" != "no"; then
...
 
   if test "$PHP_CASSANDRA" != "yes"; then
-    if test -f $PHP_CASSANDRA/include/cassandra.h; then
+    if test -f $PHP_CASSANDRA/include/cassandra/cassandra.h; then
       CPP_DRIVER_DIR=$PHP_CASSANDRA
     fi
   else
     for i in /usr/local /usr; do
-      if test -f $i/include/cassandra.h; then
+      if test -f $i/include/cassandra/cassandra.h; then
```
joeyhub, line 275: This needs some if then else for handling either situation (see remi merge).

joeyhub:
I'm not sure if I'm going to finish this. I've stripped Cassandra from my tech stack entirely as it wasn't the appropriate solution to the problem so I'll no longer need to maintain this.

mpenick:
Good catch.

PR #143, opened 2020-03-22
ulong typedef has been removed in PHP 7.4

ttraeger:
see https://raw.githubusercontent.com/php/php-src/PHP-7.4/UPGRADING.INTERNALS

### DIFF
--- a/ext/php_driver.h
+++ b/ext/php_driver.h
@@ -117,6 +117,9 @@ typedef zend_object php5to7_zend_object_free;
 typedef zval **php5to7_zval_gc;
 typedef zval *php5to7_dtor;
 typedef size_t php5to7_size;
+#if ((PHP_MAJOR_VERSION == 7 && PHP_MINOR_VERSION >= 4) || PHP_MAJOR_VERSION > 7)
+  typedef unsigned long ulong;
+#endif
 
 static inline int
 php5to7_string_compare(php5to7_string s1, php5to7_string s2)


stefnats:
I stumble upon this bug as well when manually compiling (which is necessary for 7.3+).
However, this PR would only fix this for PHP `^7.4` and not 8.0 or above.
Why not change line 120 to

```
#if ((PHP_MAJOR_VERSION == 7 && PHP_MINOR_VERSION >= 4) || PHP_MAJOR_VERSION > 7)
```

ext/php_driver.h
```diff
@@ -117,6 +117,9 @@ typedef zend_object php5to7_zend_object_free;
 typedef zval **php5to7_zval_gc;
 typedef zval *php5to7_dtor;
 typedef size_t php5to7_size;
+#if PHP_MAJOR_VERSION >= 7 && PHP_MINOR_VERSION >= 4
```
stefnats: Thanks for the fix! I stumble upon this bug as well when manually compiling (which is necessary for 7.3+).
However, this PR would only fix this for PHP ^7.4 and not 8.0 or above.
Why not change line 120 to
```
#if ((PHP_MAJOR_VERSION == 7 && PHP_MINOR_VERSION >= 4) || PHP_MAJOR_VERSION > 7)
```

ttraeger: Thanks for the hint, I changed the PR as you suggested.

mpenick:
Thanks for the PR!

stefnats:
@mpenick currently installing the php cassandra extension is a real pain. When do you think will you tag a new release? Thanks in advance

adutra:
@stefnats We are preparing a new release, should take us a few weeks. Stay tuned!

eliasmajic:
@adutra curious if there is an update

adutra:
@eliasmajic unfortunately, as I said earlier today on the mailing list, we are not anymore in capacity of resuming the development on the C++ and PHP drivers.

I know how this must be disappointing for you all, but that's the way it is right now. If you can build the driver on your own, that would be your best option currently.

